PR #30, opened 2015-01-12
Limitation of default access rights to improve security

sbrefort:
Socket and localhost (IPv4 and IPv6) are set to trust. All users (except postgres) connecting from localhost and socket can authenticate without password.

This pull request sets more restrictive default methods, requesting password authentication for localhost and peer for socket. Permissive access can be granted by pillar.


nmadhok:
@sbrefort Excellent catch! Thanks for the fix!

PR #31, opened 2015-01-25
changed pg_hba.conf management-state precondition

mriedmann:
For me the default behavior of this formula is a bit strange. 
It took me some time to see that the pg_hba-file is not managed if no "pg_hba.conf"-item is given. 

I would suggest to change this to check if ACLs exists or at least write a note into the docs.

PS: Thanks for this Formula, it (after all) saved me a lot of time.

### DIFF
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -55,7 +55,7 @@ postgresql-conf:
        - service: postgresql
 {% endif %}
 
-{% if 'pg_hba.conf' in pillar.get('postgres', {}) %}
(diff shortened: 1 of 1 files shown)


gravyboat:
Looks like we missed this one or didn't discuss it enough. Do you still find this to be an issue in the latest revision?

vutny:
I guess this is no longer applicable to the latest version of the formula.
Currently, `pg_hba.conf` file is always provisioned with default values without any conditions to check.

PR #32, opened 2015-01-29
Option to disable installation of pkg_libpq_dev

brot:


babilen:
Why the explicit comparison against `False` ? An `if True` would suffice, wouldn't it?

brot:
That wasn't my idea. I used the same pattern that already exists in this file.
Look at: https://github.com/saltstack-formulas/postgres-formula/blob/master/postgres/init.sls

gravyboat:
@brot @babilen I think there was actually some reason that the != False was used, if you look through previous PRs you can probably find it, but either myself, or someone else asked about that.

nmadhok:
Merging this. Remember seeing this in another PR with justified reason. Will search and see if I can find the reason.

PR #34, opened 2015-02-01
Add option to use upstream repo on Debian

alanpearce:
This allows Debian users to use the official PostgreSQL repository, allowing the installation of versions not provided by their distribution repository.

Note that the _postgres_ package just pulls in the latest version, which is why I added the suffix to `pkg`. Otherwise `9.4` would be installed and the formula would expect its configuration in the `9.1` folder (on _wheezy_ anyway). To actually use a different version, `postgres.lookup.id` needs to be set to `'9.4'` or similar. Otherwise the formula will just install the distribution-default version.

Should be pretty easy to adapt for other APT/RPM distributions (that's all `salt.states.pkgrepo` supports) by following the [PostgreSQL distribution-specific instructions](http://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkgrepo.html), but I don't have any such boxen to test with.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -1,6 +1,8 @@
 postgres:
   pg_hba.conf: salt://postgres/pg_hba.conf
 
+  use_upstream_repo: True
+
   lookup:
     pkg: 'postgresql-9.3'
     pg_hba: '/etc/postgresql/9.3/main/pg_hba.conf'
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -1,8 +1,21 @@
 {% from "postgres/map.jinja" import postgres with context %}
 
+{% if 'use_upstream_repo' in pillar.get('postgres') %}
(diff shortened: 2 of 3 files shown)


alanpearce:
This isn't ready just yet. I tried it on a fresh box and `service postgres initdb` isn't implemented. Hmm. I guess I should look at how debian sets it up and try to mimic that (at least for debian…)

nmadhok:
@alanpearce Will wait for further changes before merging.

alanpearce:
Turns out the problem wasn't what I thought. I hadn't generated the system locale I'd provisioned, so the `initdb` call was erroring. Once I fixed that error in my states, I was able to spin up a new VM and get Postgres running on it first time.

So it was ready all along! :smile: :shipit:

gravyboat:
Good stuff, thanks @alanpearce!

PR #38, opened 2015-02-23
Move pg_hba.conf to a standard template path

masterkorp:
It uses the standard ``templates/`` path to store the templates and the extension of the template engine used.

Beware that this is a breaking change. I will break current setups that set use``pg_hba.conf`` pillar setting to the default path.
Implementing [semver](http://semver.org/) would be advisable right now.


iggy:
Or you could just not make a backward incompatible change unless you really need to... What's the idea behind this? Do you need this change to fix something else or are you just trying to make it uniform?

masterkorp:
@iggy  the objective is try and make it uniform and the directory tree cleaner as we add more templates into the formula.

iggy:
I'd say put it in a TODO list somewhere (either in the README file or a separate TODO file) and just make a note that this should be cleaned up when someone is making more invasive, backward-incompatible changes. Breaking backwards-compatibility just for the sake of uniformity is not a great idea.

Or find a way to do it that doesn't break backwards-compatibility (and isn't clunky in it's own right).

gravyboat:
Most of the formulas are using a `files` directory for their storage. This is also the standard used in the docs. I believe this formula is this way because it's pretty old. At some point this should probably have a total rewrite/restructure.

masterkorp:
@gravyboat Would now be a better time than later ?

I am willing to open the pull request for it. But we would merge them to a separated branch while we work all the necessary changes, how about that ?

gravyboat:
@masterkorp I just took a look at the formula and it looks fine minus the somewhat weird lack of files for the structure stuff so please forgive my earlier comment. I'd be willing to merge in that change (creation of the file dir and movement of the file there to meet standards), @iggy are you okay with that?

iggy:
As long as it doesn't break backwards compatibility for an aesthetic change, I'm okay with it. If you can't do it in a backwards compatible way, save it for later.

postgres/init.sls
```diff
@@ -81,7 +81,7 @@ postgresql-conf:
 pg_hba.conf:
   file.managed:
     - name: {{ postgres.conf_dir }}/pg_hba.conf
-    - source: {{ salt['pillar.get']('postgres:pg_hba.conf', 'salt://postgres/pg_hba.conf') }}
+    - source: {{ salt['pillar.get']('postgres:pg_hba.conf', 'salt://postgres/templates/pg_hba.conf') }}
```
iggy: pg_hba.conf.jinja ?

pillar.example
```diff
@@ -1,5 +1,5 @@
 postgres:
-  pg_hba.conf: salt://postgres/pg_hba.conf
+  pg_hba.conf: salt://postgres/templates/pg_hba.conf.jinja
```
iggy: Standard -formula practice is to put files in a files/ dir (vs templates/). See... every other formula that has files (collectd, rsyslog, graphite, etc). So far changing the actual extension of the file seems to be up to the author (there's no consensus, but more often than not it's not .jinja or whatever).

iggy:
I just went and looked at the formula's use of that pillar setting and it seems really odd. I'd say the formula should just be changed to install a pg_hba.conf by default (instead of only doing it if the pillar is set). It should also be updated to install either a common sane pg_hba.conf or go and dig out the different versions for the supported packages (if a common one isn't doable). As part of that, move the file and you don't really have to worry about backwards compatibility because you aren't keying off that anymore. Thoughts?

gravyboat:
@iggy The version of postgres would need to be considered as well to ensure the directory is set up. I don't think a standard pg_hba.conf should exist, instead it should do the standard thing where it's populated from the pillar data. Maybe we could do some defaults if that doesn't exist? I feel like the 'save it for later' thing is kind of going to screw us down the line. How long do we wait to move it to a standard location? 6 months? 10 months? I'm not saying I have a solution, just not sure how waiting to move it helps us at all.

iggy:
Wait till we make some other backward incompatible change that isn't purely aesthetic.

masterkorp:
@gravyboat :+1: 

I think waiting is a bit of a waste of time, we can keep working on the side and them even when master breaks this functionality we would be an old version.

Implementing semver on the formula would also help on this migration.

iggy:
I still think breaking backwards compatibility for aesthetic reasons is ridiculous. You'd be fired if you worked here...

How about move it and leave a symlink for a while (and a big note in the commit message, README, and pillar.example that mentions what's going on)?

masterkorp:
@iggy I think you misunderstood me, I completely agree with you.

My suggestion is to create another branch named ``release-2.0.0`` (or anything that makes sense) and I will open the pull request against it.
This will allow me to open more pull requests that depend on this aesthetic change without making anyone life a nightmare.

iggy:
Formulas aren't versioned (and likely won't be for some time to come). If you want to work on more extensive changes, you fork the repo. Develop on your own fork, then when you're ready do a PR. There's no need to have branches on the upstream formulas. Not to mention that could play hell with people using gitfs (pretty much anyone using formulas) as that would auto-generate an environment for each branch.

gravyboat:
@iggy If someone is pulling directly from the master branch of the formula repo that is their own problem. If they decided not to fork and an update to bring things in line with the standard breaks that is on them. I still say we should be making changes which bring items up to the standards expected of a formula, even if it is just moving a file to the correct directory.

iggy:
We can argue our own view points back and forth till hell freezes over. Do whatever you want. I could care less. Put notices in the commit message, docs, pillar.example, and anywhere else, or just let it be a surprise to everyone that hits it.

nmadhok:
Closing this pull request. This is a big change and we don't want to break things for users.

PR #39, opened 2015-02-23
Update README to markdown syntax

masterkorp:


babilen:
Ta!

nmadhok:
@masterkorp I do not see a problem in using Restructured Text format for the Readme. Also, other formulas are using this same format. Is there a reason why you want to change it to markdown? If not then I'm going to close this one.

nmadhok:
@babilen just noticed you've just merged this. This does not follow the syntax followed in various other salt formulas.

babilen:
@nmadhok The reasoning was one of uniformity and, at least in the context of GH, markdown is much more commonly used. Which syntax are you referring to exactly?

nmadhok:
@babilen The template for salt formulas as well as other salt formulas use restructured text as the syntax instead of markdown. I feel this change breaks the uniformity with respect to other salt formulas

babilen:
@nmadhok Fair enough, that is actually a very good argument in favour of reverting this. If I were to decide a standard format I would have picked Md as it is more common in the context of Github, but RST is the dominant format in the Python community. I guess that latter explains why it is being used.

masterkorp:
I am in favour of a standard.

It seemed to me that markdown is the most easily readable.

On whatever format we take. I think we should make it a norm and follow it.

nmadhok:
@babilen markdown is a good text formatting syntax but since we're already using RST in all formulas, I would be in favor of reverting this. I'll go ahead and revert if that's alright.

gravyboat:
@babilen @nmadhok @masterkorp Another good reason to keep RST as the standard for the readmes is that all the docs are written in RST as well, so it makes it easy for people to flip back and forth between them.

Issue #41, opened 2015-02-26
Incorrect behavor using 'use_upstream_repo'


Niyakiy:
hi, there is issue with 'use_upstream_repo' pillar variable:
this time it used like:
`
{% if 'use_upstream_repo' in pillar.get('postgres') %}
install-postgresql-repo:
 .... omit some strings ....
{% endif %}
`
but such usage checks only existence of variable in pillar but not its value.
looks like it should be something like this:

`{% if 'use_upstream_repo' in pillar.get('postgres') %}
{% if salt['pillar.get']('postgres:use_upstream_repo', False) %}
install-postgresql-repo:
 .... omit some strings ....
{% endif %}
{% endif %} `

iggy:
Protip: Use 3 ` (backticks) around code to keep markdown from trying to interpret parts of it. 

I agree. Feel free to submit a PR to change it.

You should be able to just do:
```
{% if salt['pillar.get']('postgres:use_upstream_repo') %}
...
{% endif %}
```

rmoorman:
Hello,

The lines mentioned in this ticket also cause errors in case no pillar data is set at all for "postgres". It should work out of the box with no additional configuration as stated in the [documentation](http://docs.saltstack.com/en/latest/topics/development/conventions/formulas.html#configuring-formula-using-pillar).

However, I get the following error:

````
[CRITICAL] Rendering SLS 'base:postgres' failed: Jinja error: argument of type 'NoneType' is not iterable
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/utils/templates.py", line 286, in render_jinja_tmpl
    output = template.render(**unicode_context)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 969, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 742, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "<template>", line 3, in top-level template code
TypeError: argument of type 'NoneType' is not iterable

; line 3

---
{% from "postgres/map.jinja" import postgres with context %}

{% if 'use_upstream_repo' in pillar.get('postgres') %}    <======================
install-postgresql-repo:
  pkgrepo.managed:
    - humanname: PostgreSQL Official Repository
    - name: {{ postgres.pkg_repo }}
[...]

PR #42, opened 2015-02-26
Adjust the init state's jinja syntax to prevent errors if there is no postgres pillar data

rmoorman:
This pull request contains changes which prevent errors if there is no postgres pillar data (as discussed in in issue #41). Please verify that this is working correctly for you too.

### DIFF
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -1,6 +1,6 @@
 {% from "postgres/map.jinja" import postgres with context %}
 
-{% if 'use_upstream_repo' in pillar.get('postgres') %}
+{% if salt['pillar.get']('postgres:use_upstream_repo') %}
 install-postgresql-repo:
   pkgrepo.managed:
     - humanname: PostgreSQL Official Repository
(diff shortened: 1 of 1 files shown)


postgres/init.sls
```diff
@@ -15,7 +15,7 @@ install-postgresql-repo:
 install-postgresql:
   pkg.installed:
     - name: {{ postgres.pkg }}
-    - refresh: {{ 'use_upstream_repo' in pillar.get('postgres') }}
+    - refresh: {{ salt['pillar.get']('postgres:use_upstream_repo') }}
```
iggy: I think you actually want to set a default here, otherwise refresh will get set to exceptions.KeyError.
```
{{ salt['pillar.get']('postgres:use_upstream_repo', False) }}
```

rmoorman: That also applies to the if condition further up I guess?

iggy: Test it to double check, but I'm pretty sure the default returned by pillar.get evaluates to False (at least enough for an if test that is).

rmoorman: I double checked it and it works without a default in the if test. Thank you very much.

iggy:
Looks good to me.

nmadhok:
Thanks @rmoorman

rmoorman:
You are welcome

PR #44, opened 2015-03-12
allow multiline postgres:postgresconf

dferrantelli:
Added automatic indentation of postgres:postgresconf to allow for multiple lines to be added to the conf as per https://github.com/saltstack/salt/pull/5920


gravyboat:
Great addition, thanks!

PR #45, opened 2015-04-21
added client-only formula and required default values

dferrantelli:


iggy:
You could move the pkgrepo bit off into an include and then extend it (to change the require_in).

gravyboat:
This is a good move towards simplifying the overloaded init. Does anyone remember if https://github.com/saltstack-formulas/postgres-formula/pull/45/files#diff-388d76ce7dc36409e9945ef2b4c11ccbR8 is a problem for non ubuntu systems to process?

iggy:
Works on Debian at least.

gravyboat:
@iggy Derivatives don't count! I mean on a non apt based system.

dferrantelli:
@iggy I like the idea of the pkgrepo in an include - i'll do that in a minute. 

@gravyboat keyid and keyserver are specific to Debian derivatives. I could add a check to make sure it's a Debian based system? the server package also uses the same code (though once i make it an include that will be a simple change for both)

gravyboat:
@dferrantelli I honestly can't remember if adding those on non-debian/ubuntu systems causes an error or not. A check would probably be good.

dferrantelli:
Now that it is an include, it should happen first as those are handled before the rest of the state file.

PR #46, opened 2015-04-22
Switch to new style map.jinja

iggy:
Move this formula to the new style map.jinja. Still need to sort out RedHat
handling, but it's not any worse than it was before (i.e. horribly broken)


gravyboat:
Looks pretty good to me. Thanks @iggy!

PR #47, opened 2015-04-23
Add _states/_modules for tablespaces and add tablespace support to formula

iggy:
Add temporary states/modules for handling tablespaces in Postgresql. These will
be fed back upstream, but they can live here for now.

Add support for the states/modules to the formula.

### DIFF
--- a/_modules/postgres_ext.py
+++ b/_modules/postgres_ext.py
@@ -1,0 +1,261 @@
+from __future__ import absolute_import
+import logging
+try:
+    import pipes
+    import csv
+    HAS_ALL_IMPORTS = True
+except ImportError:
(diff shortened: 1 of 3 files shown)


gravyboat:
Hmm, should the tablespace stuff live in the init?

iggy:
Everything else does (database and user creation)... I actually just copied the user creation from a few lines above that to get started.

gravyboat:
Okay cool I will merge this then. Thanks @iggy.

iggy:
I do have plans to break this thing up a little bit into separate states. But I really needed to get this stuff in there for our maintenance window at work next week.

Issue #50, opened 2015-04-27
tablespaces missing from example


jmdcal:
it looks like tablespaces are missing from the example pillar but required by the template

iggy:
They are in default.yaml (as of #49 ). I haven't added them to the pillar.example yet because I think I'm going to change the layout of the dict. When I first did the tablespace support all I supported was name/directory. When I did this round of ts support, I actually enabled the options/owner/etc support (but my original pillar structure didn't support that).

vutny:
The `pillar.example` file already has `tablespaces` item. Looks like this is no longer an issue.

PR #51, opened 2015-04-27
namespaces are not mandatory

jmdcal:
don't make namespaces mandatory


iggy:
Unnecessary as explained in #50.

PR #56, opened 2015-05-06
added pkg_repo defaults for trusty and utopic

dferrantelli:


gravyboat:
Is this just for examples? I don't see how this is actually imported.

babilen:
@gravyboat It is being imported in `map.jinja` AFAICT.

PR #57, opened 2015-05-08
Added superuser pillar, missing conf_dir directory

renoirb:


postgres/init.sls
```diff
@@ -5,6 +5,10 @@ include:
   - postgres.upstream
 {% endif %}
 
+{{ postgres.conf_dir }}:
+  file.directory:
+    - makedirs: True
```
iggy: I think you could get away with just putting `- makedirs: True` on all the stanzas that actually touch the config files... Although I suspect if the directory doesn't exist, you've got some other bug (package didn't install, etc.)

PR #61, opened 2015-07-10
Fixes for debian specific configuration

madssj:
There might be a better way of implementing this, but I fail to see it.

I'll be happy to fix it another way, but would like some guidance on how, if need be.


iggy:
This looks right, but I feel like I handled this another way. Let me check my copy later today.

puneetk:
@iggy send over a pull when you fix it the other way

PR #68, opened 2015-11-06
Bugfix/fixing require statements

blast-hardcheese:
- Converting all pillar references to their actual state names
```
  - require:
      - pkg: {{ postgres.pkg }}
```
should actually be
```
  - require:
      - pkg: install-postgresql
```

- Made `postgres-db-{{ name }}` actually require `run-postgres`
- Added an override for `initdb`, as `service` provided in CentOS 7 is just a shim for systemctl, and doesn't support custom commands.


gravyboat:
Thanks!

PR #69, opened 2015-11-15
Moving initdb command definition

blast-hardcheese:
Fix for the previously incomplete PR-- I mistakenly thought that Debian-based distros were the only ones that would continue to support `service postgresql initdb`. @iggy's point is totally valid, for backwards compatibility though, so I'm moving it into `defaults`. Also wrapping things up by adding an example to `pillar.example`.


blast-hardcheese:
Thanks @gravyboat!

PR #73, opened 2015-11-23
ensure proper permissions for postgresql dir

rominf:


postgres/init.sls
```diff
@@ -8,6 +8,11 @@ include:
 {{ postgres.conf_dir }}:
   file.directory:
     - makedirs: True
+    - user: postgres
+    - group: postgres
+    - dir_mode: 700
+    - clean: True
```
iggy: This is probably a (backwards incompatible) change that should be discussed with the user base.

puneetk: you are right, let me send a pull to remove the clean:True

rominf: It test's for `PG_VERSION` file. So if the base is initialized, it won't be removed. But if this dir contains something else, it would be cleared what is good. If you want to avoid this behaviour, add `postgres.clean_conf_dir` variable.

blbradley: This change also causes this:

```
          ID: /etc/postgresql/9.4/main
    Function: file.directory
      Result: False
     Comment: User postgres is not available Group postgres is not available
     Started: 08:54:51.746521
    Duration: 63.534 ms
     Changes:
```

Please check your changes for idempotency before making commits and (more importantly) submitting your PR. This means they should work on a fresh box with no problems.

rominf: @blbradley, agreed. My fault, sorry.

The most reliable way to test Salt modules is to write tests with `unittest`, but sometimes it's tedious to write them, so it's easier test Salt by hands. But I want exhaustive testing, not just in one "special" OS.
How do you test modules? Are there Docker/Vagrant/etc. containers for testing Salt states in automated way?

blbradley: @rominf Do you have some documentation about testing formulas with `unittest`?

This module looks like it now works with [test-kitchen](http://kitchen.ci/). You may want to look at that for testing.

gravyboat:
In what scenario would you add the conf dir prior to the package installation?

postgres/init.sls
rominf: @blbradley I think that it's possible to put formulas inside Salt sources and then use this: https://docs.saltstack.com/en/latest/topics/development/tests/unit.html

Thank you for kitchen.ci, I'll definitely check it out.

rominf:
@gravyboat 
I've changed this because it didn't work for openSUSE 13.1 with already installed PostgreSQL. Conf directory not always (it depends on package manager and flags) gets cleared on removing `postgres` package.

gravyboat:
@rominf Gotcha, @blbradley and @rominf  What do you guys think about closing #78, and instead we modify what was added to only be used for opensuse via a grain check? Any opinions on something like that, or other pitfalls we could potentially fall into that would negatively impact users depending on their OS?

rominf:
@gravyboat 
I have different idea:
```yaml
{{ postgres.conf_dir }}:
  file.directory:
    - makedirs: True
{% if 'postgres' in salt['user.list_users']() and 'postgres' in salt['user.list_groups']() %}
    - user: postgres
    - group: postgres
    - dir_mode: 700
    - clean: True
    - unless: test -f {{ postgres.conf_dir }}/PG_VERSION
{% endif %}
```
Not sure about `dir_mode`, `clean`inside `if` though.

blbradley:
@rominf Why does the config directory need to be cleared when removing the package?

rominf:
@blbradley 
Normally you don't need to clear configuration. But `postgres-formula` didn't work with existing config directory, that's why I added `file.absent` to `postgres.dropped` state and `clean=True` (if it's not a working installation and `PG_VERSION` file is absent).

blbradley:
@rominf  Config files are preserved on Debian unless `apt-get purge` is used. Maybe what you were wanting was `pkg.purged`.

@gravyboat Probably recommend using `pkg.purged` in `postgres/dropped.sls` after removing code from this PR.

rominf:
@blbradley, you are right, `pkg.purged` is what I needed.

blbradley:
@rominf @gravyboat Hello again. Are we going to remove this or do something else?

gravyboat:
@blbradley Do you just want to prep a PR to back this out?

rominf:
@blbradley, @gravyboat I'm planning to rewrite this place and add something else, TEST IT :), but later. I agree with @gravyboat, that the easiest thing is to backout this PR, moreover it's self-contained.

blbradley:
@rominf If you are currently working on this, you are more than welcome to back this out. I might be a few days.

blbradley:
@gravyboat @puneetk Can we revert #77 and this PR in that order?

puneetk:
@blbradley  it would be easier to make a new pr with the changes removed, can you send over a pull with the changes fixed i can merge it in. If not, i can do the same.

blbradley:
@puneetk Go ahead!

PR #74, opened 2015-11-23
replace pkg_contrib with pkgs_extra

rominf:


iggy:
This is a backwards incompatible change.

rominf:
Yes, I know. I changed this because `postgres-contrib` is not very different from `postgres-plpython` and several similar packages.

iggy:
It's still a backwards incompatible change. Either add `postgresql-contrib` back in (and gate it if you really want to) or revert this PR.

PR #78, opened 2015-12-06
add 'postgres' user and group

blbradley:
Fixes idempotency on Debian. Can someone verify on CentOS? I may be able to with more work.

### DIFF
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -5,6 +5,12 @@ include:
   - postgres.upstream
 {% endif %}
 
+postgres:
+  user.present:
+    - system: True
+    - shell: /bin/bash
+    - home: /var/lib/postgres
+
 {{ postgres.conf_dir }}:
   file.directory:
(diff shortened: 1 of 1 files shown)


gravyboat:
The postgres user should be created when the package is installed, is that not the case for your use case?

blbradley:
Yes, it is. But please see #73. There is now a `file.directory` state that runs before the `pkg.installed` state that requires the user/group.

gravyboat:
Okay, asking a question over there.

blbradley:
@gravyboat To be clear, I fully agree that the user and group creation should be handled by packages.

puneetk:
@gravyboat this looks like a 'fix' for the other pull, merge away ?

gravyboat:
@puneetk Please don't merge this. We're going to back out the other PR to get this fixed.

blbradley:
@puneetk @gravyboat Agreed!

PR #79, opened 2015-12-16
Added support for pg_ident.conf

kresike:
I needed the formula to handle pg_ident.conf, so I added it. Maybe someone else finds this useful.


gravyboat:
Hmm, should this be wrapped in an if statement to only take effect if this section of the pillar exists? Seems like it might cause confusion if you missed this.

noelmcloughlin:
Superseded by #181 @evasdk, fyi.

PR #86, opened 2016-03-14
Added option to disable creation of backup file when postgresql.conf …

campbellmc:
Added postgresconf_backup option which allows disabling of .bak file creation for postgresql.conf when it changes.
Due to fopen issue with file.blockreplace - .bak file is currently created with root:root ownership - which causes downstream issues with wal-e backup-push and messes around with IDS.
Default behaviour doesn't change - file is backed up unless explicitly configured otherwise.

### DIFF
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -59,6 +59,9 @@ postgresql-conf:
         {{ postgres.postgresconf|indent(8) }}
     - show_changes: True
     - append_if_not_found: True
+    {% if not postgres.postgresconf_backup|default(True) -%}
+    - backup: False
+    {% endif -%}
     - watch_in:
        - service: run-postgresql
 {% endif %}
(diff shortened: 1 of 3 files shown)


gravyboat:
Good catch, thanks!

postgres/init.sls
```diff
@@ -59,6 +59,9 @@ postgresql-conf:
         {{ postgres.postgresconf|indent(8) }}
     - show_changes: True
     - append_if_not_found: True
+    {% if not postgres.postgresconf_backup|default(True) -%}
```
iggy: Just out of curiosity, was the `default()` necessary here? Did you need it to fix something or did you just add it out of habit/whatever?

campbellmc: Habit. There is no real need for it as the default is already provided.  I will remove it at some stage.

PR #87, opened 2016-03-16
The user must be postgres when run initdb

aldovillagra:


iggy:
This seems like a change worthy of seeing what kind of testing you did and/or what issues you were seeing before.

aldovillagra:
please run ls -la {{ postgres.conf_dir }}, {{ postgres.data_dir }} you will see:

- postgres user is the owner of data files.
- postgres user is the owner of config dir.

iggy:
Is this actually fixing a problem you were seeing? Because if there were an issue, it seems like it would have cropped up before. And I used to actually use this formula without issue.

tgoodaire:
This breaks FreeBSD support because FreeBSD uses the pgsql user and group, not postgres. 

ID: pg_hba.conf
Function: file.managed
Name: /var/lib/pgsql/data/pg_hba.conf
Result: False
Comment: User postgres is not available Group postgres is not available
Started: 15:17:40.900433
Duration: 0.958 ms

tgoodaire:
I propose a change like this instead. I created two new variables - postgres_user and postgres_group. They default to "postgres" but may be set to some other value in the pillar. This same problem with hard-coding the postgres username also breaks managing databases in database users, so this ought to fix that too.

`diff --git a/postgres/defaults.yaml b/postgres/defaults.yaml
index 2d0013a..25a32a2 100644
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -20,3 +20,5 @@ postgres:
   pg_hba.conf: salt://postgres/pg_hba.conf
   commands:
     initdb: service postgresql initdb
+  postgres_user: postgres
+  postgres_group: postgres
diff --git a/postgres/init.sls b/postgres/init.sls
index 74d44e9..076b68e 100644
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -71,8 +71,8 @@ pg_hba.conf:
     - name: {{ postgres.conf_dir }}/pg_hba.conf
     - source: {{ postgres['pg_hba.conf'] }}
     - template: jinja
-    - user: postgres
-    - group: postgres
+    - user: {{ postgres.postgres_user }}
+    - group: {{ postgres.postgres_group }}
     - mode: 644
     - require:
       - pkg: install-postgresql
@@ -90,14 +90,14 @@ postgres-user-{{ name }}:
     - inherit: {{ user.get('inherit', True) }}
     - replication: {{ user.get('replication', False) }}
     - password: {{ user.get('password', 'changethis') }}
-    - user: {{ user.get('runas', 'postgres') }}
+    - user: {{ user.get('runas', postgres.postgres_user) }}
     - superuser: {{ user.get('superuser', False) }}
     - require:
       - service: run-postgresql
 {% else %}
   postgres_user.absent:
     - name: {{ name }}
-    - user: {{ user.get('runas', 'postgres') }}
+    - user: {{ user.get('runas', postgres.postgres_user) }}
     - require:
       - service: run-postgresql
 {% endif %}
@@ -114,7 +114,7 @@ postgres-db-{{ name }}:
     {% if db.get('owner') %}
     - owner: {{ db.get('owner') }}
     {% endif %}
-    - user: {{ db.get('runas', 'postgres') }}
+    - user: {{ db.get('runas', postgres.postgres_user) }}
     - require:
         - service: run-postgresql
     {% if db.get('user') %}
@@ -140,7 +140,7 @@ postgres-schema-{{ schema }}-for-db-{{ name }}:
 postgres-ext-{{ ext }}-for-db-{{ name }}:
   postgres_extension.present:
     - name: {{ ext }}
-    - user: {{ db.get('runas', 'postgres') }}
+    - user: {{ db.get('runas', postgres.postgres_user) }}
     - maintenance_db: {{ name }}
 {% if ext_args is not none %}
 {% for arg, value in ext_args.items() %}

`

tgoodaire:
This issue should be resolved by https://github.com/saltstack-formulas/postgres-formula/pull/90 which has been merged.

babilen:
Shall we close it, @iggy, @aldovillagra ?

aboe76:
I Have tested this formula on archlinux and the initdb cmd will still be ran as root instead of postgres user...

gravyboat:
@aboe76 Did you test the change from #90?

aboe76:
Yes just pulled the latest formula and tested on archlinux only change to the formula was the above 

change user: root into user: {{ postgres.postgres_user }}

gravyboat:
@aboe76 are you talking about this one? https://github.com/tgoodaire/postgres-formula/blob/master/postgres/init.sls#L23 or which one specifically are you having issues with from #90? I'll tag @tgoodaire as well so we can get this resolved.

aboe76:
@gravyboat , No I was referencing:

https://github.com/saltstack-formulas/postgres-formula/pull/87/files#diff-cc80745600f12cbaecc09eed76b8034bR34

gravyboat:
@aboe76 Gotcha, can you try with the changes from 90?

tgoodaire:
@aboe76 - You may try the changes in https://github.com/saltstack-formulas/postgres-formula/pull/93
One of the changes is the addition of an "initdb_user", which in your case could be set to "postgres". Most distros use a service or init script that is intended to be run by root, but will initialize the db cluster as the appropriate postgres user.

aboe76:
this one can close with because of #93

PR #88, opened 2016-05-01
Add Xenial to code name map

mattysads:


gravyboat:
Good stuff, thanks Sadler!

PR #92, opened 2016-05-25
Reload postgresql server instead of restarting it

### LINKED ISSUE #91: postgres restarts on highstate - intended behaviour?
we've started using the postgres formula and encountered an issue when highstating. The database server does a restart instead of a reload which leads to problems with apps using the database, which also need a restart.

I guess that the reason is inside of the [run-postgresql state](https://github.com/saltstack-formulas/postgres-formula/blob/e9e41eb12c22a3cb487a0063195444f7b84e184b/postgres/init.sls#L43).

```yaml
run-postgresql:
  service.running:
    - enable: true
    - require:
      - pkg: install-postgresql
```

Are there any things done by the formula which require a restart of the service, or may this enhanced by `reload: True`?

dahrens:
Fixes #91

In case of changes in postgresql.conf and pg_hba.conf the
postgresql will now reload instead of restart.

### DIFF
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -40,7 +40,8 @@ postgresql-initdb:
 
 run-postgresql:
   service.running:
-    - enable: true
+    - enable: True
+    - reload: True
     - name: {{ postgres.service }}
     - require:
       - pkg: install-postgresql


gravyboat:
Looks good to me, thanks.

Issue #81, opened 2016-01-04
First state run fails


EvaSDK:
Here is the output of state.highstate:
```
----------
          ID: /etc/postgresql/9.4/main
    Function: file.directory
      Result: False
     Comment: User postgres is not available Group postgres is not available
     Started: 15:54:34.472581
    Duration: 428.798 ms
     Changes:   
----------
```

The issue is that the following block is not requiring postgres to be already installed so the file does not exist yet:
https://github.com/saltstack-formulas/postgres-formula/blob/master/postgres/init.sls#L56

EvaSDK:
Actually this block: https://github.com/saltstack-formulas/postgres-formula/blob/master/postgres/init.sls#L8

but it appears there is a dependency problem for other files in this formula.

vutny:
I guess this is already resolved and should be closed.

EvaSDK:
I recently reinstalled some VMs with this state and it is indeed fixed afaict. Thanks.

PR #99, opened 2016-06-05
Adding pkg dependency for directory permissions.

a-powell:
Fixes states failing on first execution of highstate (https://github.com/saltstack-formulas/postgres-formula/issues/81).


babilen:
Thanks :)

Issue #100, opened 2016-06-06
postgresql-cluster-prepared state is broken


EvaSDK:
Current logic for postgresql-cluster-prepared state does not work on Debian OS family and other OSes most likely due to initdb and create_cluster being false by default. The flaws is that the id cover two boolean that create 4 different states at least 2 of which are invalid (both true and both false).

I will submit a PR to clean that up in a few minutes.

aboe76:
@EvaSDK I'm sorry to say but there isn't a pg_createcluster on Archlinux so your modification will fail on Archlinux, and I couldn't find anything usefull in the postgres source code,
so this looks like a ubuntu / debian specific command...

Can you confirm this?

```
local:
----------
          ID: postgresql-cluster-prepared
    Function: cmd.run
        Name: test -f /var/lib/postgres/data/environment
      Result: False
     Comment: Command "test -f /var/lib/postgres/data/environment" run
     Started: 21:44:11.121038
    Duration: 63.02 ms
     Changes:   
              ----------
              pid:
                  3277
              retcode:
                  1
              stderr:
              stdout:

Summary for local
```

EvaSDK:
I didn't add that command.

EvaSDK:
Ok so I got tricked by the formula into thinking initdb is actually createdb so my fix is indeed wrong but so was the formula anyway. https://www.postgresql.org/docs/9.4/static/creating-cluster.html

aboe76:
@EvaSDK Thanks for looking into this, can you create a pr? or should I

EvaSDK:
I am working on it right now.

EvaSDK:
Fixed.

PR #101, opened 2016-06-06
Debian family fixes

EvaSDK:
Fix an issue with 2 pillar values controlling the content of the same state id, see #100.
Update codenamemap with Ubuntu wily release (which is my current production servers) and general updates to Debian family OSes logic/codenamemap settings.


aboe76:
thanks for this

PR #102, opened 2016-06-07
Fix broken Arch initdb from merge of PR#101

EvaSDK:
As discussed in #100, the state fooled me into thinking initdb created the database while it takes care of the cluster. The state provided two settings to select the method to create the cluster. I think it should be controlled by just one and since Debian has its specific tools just reflect that in the state.

If an admin of a Debian server wants to use initdb directly, it can still be done by setting `initdb: True` in `postgres:lookup`.

Also include a minor fix for `osmap` usage of Arch and a slightly less verbose `pkg_dev` setting for Debian family OS.


aboe76:
@EvaSDK thanks looks good, but will test it first at home on my arch box before merging if you don't mind.

EvaSDK:
No problem.

aboe76:
Works great with the changes,

PR #103, opened 2016-06-08
Allow to specify a tablespace for a database

javierbertoli:
* Tablespaces have to be created before the databases
* Add pillar examples
* Add tests

### DIFF
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -112,6 +112,27 @@ postgresql-user-{{ name }}:
       - service: postgresql-running
 {% endfor %}
 
+{% for name, directory in postgres.tablespaces.items()  %}
+postgresql-tablespace-dir-perms-{{ directory}}:
+  file.directory:
+    - name: {{ directory }}
+    - user: postgres
+    - group: postgres
+    - makedirs: True
+    - recurse:
+      - user
+      - group
+
+postgresql-tablespace-{{ name }}:
+  postgres_tablespace.present:
+    - name: {{ name }}
+    - directory: {{ directory }}
+    - user: postgres
+    - require:
+      - service: postgresql-running
+      - file: postgresql-tablespace-dir-perms-{{ directory}}
+{% endfor %}
+
 {% for name, db in postgres.databases.items()  %}
 postgresql-db-{{ name }}:
 {% if db.get('ensure', 'present') == 'absent' %}
@@ -126,6 +147,7 @@ postgresql-db-{{ name }}:
     - lc_ctype: {{ db.get('lc_ctype', 'en_US.UTF8') }}
     - lc_collate: {{ db.get('lc_collate', 'en_US.UTF8') }}
     - template: {{ db.get('template', 'template0') }}
+    - tablespace: {{ db.get('tablespace', 'pg_default') }}
     {% if db.get('owner') %}
     - owner: {{ db.get('owner') }}
     {% endif %}
@@ -169,26 +191,5 @@ postgresql-ext-{{ ext }}-for-db-{{ name }}:
 {% endfor %}
 {% endif %}
 {% endif %}
-{% endfor %}
-
-
-{% for name, directory in postgres.tablespaces.items()  %}
-postgresql-tablespace-dir-perms-{{ directory}}:
-  file.directory:
-    - name: {{ directory }}
-    - user: postgres
-    - group: postgres
-    - makedirs: True
-    - recurse:
-      - user
-      - group
-
-postgresql-tablespace-{{ name }}:
-  postgres_tablespace.present:
-    - name: {{ name }}
-    - directory: {{ directory }}
-    - require:
-      - service: postgresql-running
-      - file: postgresql-tablespace-dir-perms-{{ directory}}
 {% endfor %}
 
--- a/pillar.example
+++ b/pillar.example
@@ -47,6 +47,9 @@ postgres:
     - ['local', 'db1', 'localUser']
     - ['host', 'db2', 'remoteUser', '123.123.0.0/24']
 
+  tablespaces:
+    my_space: /srv/my_tablespace
+
   databases:
     db1:
       owner: 'localUser'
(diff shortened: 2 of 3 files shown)


aboe76:
@javierbertoli are tablespaces now mandatory?

Because of: https://github.com/saltstack-formulas/postgres-formula/pull/51

javierbertoli:
No, they're not, but a database is always created in a tablespace. If none is specified, the database is created in the tablespace named *pg_default*. 

This PR allows you to specify a tablespace if you want to, or creates the database in the default one otherwise.

aboe76:
@javierbertoli thanks for the explanation, pg_default is postgres own tablespace on template0, so I see no problems anymore.

postgres/init.sls
```diff
@@ -126,6 +147,7 @@ postgresql-db-{{ name }}:
     - lc_ctype: {{ db.get('lc_ctype', 'en_US.UTF8') }}
     - lc_collate: {{ db.get('lc_collate', 'en_US.UTF8') }}
     - template: {{ db.get('template', 'template0') }}
+    - tablespace: {{ db.get('tablespace', 'pg_default') }}
```
dynjnelson: Shouldn't this also require a:

{% if db.get('tablespace', 'pg_default') != 'pg_default' %}
   postgresql-tablespace-{{ name }}
{% endif %}

in the requirement section?
(or, alternately, a requires_in added to the tablespace state for the database?)

 +    - tablespace: {{ db.get('tablespace', 'pg_default') }}

javierbertoli: I didn't want to force the relation tablespace/database in any of the ways you mention because:

* In case you want to create a tablespace where lusers can create and/or manage databases at will, you don't have database/s to force a `require_in`, because you don't know them beforehand.

* Tablespaces can be created beforehand (ie, by some other tool/package you installed). In case you want to create databases in these tablespaces, you can just specify it in the database creation state, but I don't think it's wise to force a require to a tablespace state because you're not managing it directly.

Maybe I'm wrong? Perhaps is there a clean way to cover these two cases, that I'm missing?

dynjnelson: Consider this *related* PR as well: 
https://github.com/saltstack-formulas/postgres-formula/pull/109

There is always going to be a disconnect between allowing users to create/drop/modify databases, users, and - yes - table spaces.  However, it is an *error* to specify a database with a tablespace that doesn't exist, therefore there needs to be a dependency in that direction. There doesn't need to be one in the reverse (`require_in`), as you note, however, and in that case you are right.

At the end of the day, this is easy to fix by placing a requires on the *database* side on the tablespace.

javierbertoli: > it is an error to specify a database with a tablespace that doesn't exist, therefore there needs to be a dependency in that direction

It **is** an error, agreed. But the way I see it, specifying a dependency on a tablespace that does not exist is no different an error than specifying a package name in your pillar that does not exist (like `package: psotgresql-sevrer`).

I also think it's OK to depend on things you are not managing with Salt (yet) or that out of your control, but can be used as a dependency for what you do. And if they do not exist, it's  OK for the state to *fail*, because in the end, you *must* know your infrastructure to automate it and that unmet dependency *shows* you there's an error in the current infrastructure (and/or what you think it should be) and does not proceed any further.

What you're doing in PR #109 is to make *mandatory* that the tablespace is managed with Salt to be usable as a dependency in the database state. And I think that might *lead* to errors instead of just showing them:

Consider the following hypothetical case (I understand it's a corner case, but possible):

 * The package installer for the app **my_db_app** creates a tablespace **mydbapp_tbsp**, where it stores data, with tablespace owner **mydbapp**. In salt, I just add a package state to install that package.
* I want, using this formula, to create a database in that tablespace for some statistics, so I add a database state with `name: statistics` and `tablespace: my_db_app`. Because *I know* what I'm installing and configuring, I know the previously added package space will add the tablespace, and all will go OK.
 * One day, new version of package **my_db_app** changes the owner of the tablespace to **someotherowner**. 

This PR will still work as expected, because the states are ordered and depedendencies are set but not enforced. 

Now, consider the same example under PR #109: you're making *mandatory* to add a tablespace state to create **mydbapp_tbsp** with owner **mydbapp**. When the package **my_db_app** is updated and it changes the owner of that tablespace to **someotherowner**, your *forced* tablespace state will *revert* it back to **mydbapp**. And break things.

Again, is a corner case. But entirely possible, as described or with variants. In those cases, I prefer to offer the possibility of adding the dependency than enforcing it.

dynjnelson:
I agree that the corner case demonstrates a situation where the dependency on database -> tablespace is undesirable. However, I would argue that the needs of the many outweigh the needs of the few. It would be a violation of the salt best practices and - certainly - a POLA (principle of least astonishment) violation to fail to link the two via a dependency.

If you were to run into a situation such as the one you describe above, I would argue it would not be appropriate to manage the state of the database via this formula /at all/ and - instead - add a separate state to manage that very specific case.

Remember: I believe that the formulas are there to manage the usual case. For example, there isn't much (any?) provision here for managing multiple versions and multiple clusters of postgresql on a given host (something I would like and use), but that should not prevent the state from being appropriate for the 99% of cases out there.

Lastly, I would like to thank you for participating in a civil manner, and I would hope that you take my own participation in the spirit in which it is offered!

javierbertoli:
>  I would argue that the needs of the many outweigh the needs of the few.

Ah! That's an argument, which I respectfully decline to discuss on a PR thread. In a bar, behind a beer or any other spirit, it would be a totally different matter which I'd even enjoy to discuss :smile: 

>  It would be a violation of the salt best practices and - certainly - a POLA (principle of least astonishment) violation to fail to link the two via a dependency.

  I am not aware that Salt BP require an explicit dependency link, sorry... Mind you point me to a link to this principle? I'm relatively new in the neighborhood :smile:

> If you were to run into a situation such as the one you describe above, I would argue it would not be appropriate to manage the state of the database via this formula /at all/ and - instead - add a separate state to manage that very specific case.

Yes, or maybe extending the formula and overriding the desired states would be also a suitable solution.

> there isn't much (any?) provision here for managing multiple versions and multiple clusters of postgresql on a given host (something I would like and use)

I wonder how difficult would it be to add an outer iterator on the different versions to use and refactor the recipes to take these versions into consideration? Nice exercise to try on a boring day, isn't it? :smile: 

> Lastly, I would like to thank you for participating in a civil manner, and I would hope that you take my own participation in the spirit in which it is offered!

It's a plessure! Your arguments make sense and mine are mostly my personal preferences regarding what I expect of a formula. If your proposal is also what Salt BPs suggest, I agree with you that those changes should be merged.

PR #104, opened 2016-06-08
Remove python support

tgoodaire:
In my opinion, the responsibilities of the postgres-formula should be to install the postgres server, client, and related packages (for example, postgres contrib modules). 

My feeling is that the formula is overstepping its boundaries by supporting the installation of python bindings, psycopg2, etc. If we do intend for this formula to continue supporting python, we should consider expanding this to also deal with ruby, perl, etc packages as well.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,11 +18,6 @@ Available states
 
 Installs the postgresql package.
 
-``postgres.python``
--------------------
-
-Installs the postgresql python module
-
 ``postgres.client``
 -------------------
 
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -4,7 +4,6 @@ postgres:
   pkg_libpq_dev: postgresql-libs
   pkg_client: postgresql-client
   pkgs_extra:
-  python: python-psycopg2
   service: postgresql
   conf_dir: /var/lib/pgsql/data
   data_dir: /var/lib/pgsql/data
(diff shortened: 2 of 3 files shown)


aboe76:
@tgoodaire please don't remove the python binding, this is necessary to create a returner to postgres..

But please feel free to add extra sls files for the ruby, perl

gravyboat:
I don't really have much of an issue with the psycopg2 include here because it's not relying on other formulas to work properly. Typically I would agree and object to this if it made other formulas mandatory, but since it doesn't I agree with @aboe76, if anything the other bindings should be supported!

dynjnelson:
I happen to agree with @tgoodaire  -- I feel as though it should be supported by a supporting formula or recipe.

aboe76:
I think there are more formulas with similar issues that @tgoodaire wrote:
- mysql-formula: mysql.python
- memcached-formula: memcached.python_memcached
- qpid-formula: qpid.python-qpid
- salt-formula: salt.gitfs.gitpython, salt.gitfs.dulwich, salt.gitfs.pygit2

The python bindings are mostly there for a reason within saltstack, to enhance a state, module or returner. The other bindings are a nice thing to have and make it easier to expand the formula within somebodies workflow / deployment. 

So please @tgoodaire create a perl.sls and a ruby.sls if you want to add those.

PR #107, opened 2016-07-08
Simplify codenamemap

tgoodaire:
Use a jinja2 macro to simplify and make more consistent settings for Ubuntu.
We believe that this is a drop in replacement and doesn't semantically change the content or behaviour of the formula.


gravyboat:
Do you remember when macro support was added for jinja2? I like this change a lot.

tgoodaire:
I'm not sure when macro support was added to jinja2. From the release history, there was a bugfix to macros in 2.1 - "fixed a bug with empty statements in macros.". So support goes at least that far back.

gravyboat:
Yeah that was all I found as well, I'm fine with merging this.

PR #109, opened 2016-07-08
fix missing dependency on tablespace for database creation and also fix db.get('owner') vs. db.get('user')

dynjnelson:
fix missing dependency on tablespace for database creation and also fix db.get('owner') vs. db.get('user')

### DIFF
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -154,8 +154,11 @@ postgresql-db-{{ name }}:
     - user: {{ db.get('runas', postgres.user) }}
     - require:
       - service: postgresql-running
-    {% if db.get('user') %}
-      - postgres_user: postgresql-user-{{ db.get('user') }}
+    {% if db.get('tablespace') %}
+      - postgres_tablespace: postgresql-tablespace-{{ name }}
+    {% endif %}
+    {% if db.get('owner') %}
+      - postgres_user: postgresql-user-{{ db.get('owner') }}
     {% endif %}
 
 {% if db.schemas is defined %}


dynjnelson:
Should this be *two* pull requests?
One to fix the dep and one to fix the use of db.get('owner') vs. db.get('user') ?

gravyboat:
It probably should, right now I'm waiting for people to weigh in. @babilen, @iggy any opinions on this? I'm a bit concerned for backwards compat.

babilen:
Even though I agree that `owner` is a better name for that option, we should think twice before breaking backwards compatibility. My suggestion would be to allow both (user and owner), but to document `owner` only so that new users start to use that.

dynjnelson:
So a colleague and I dug into this a bit deeper.
In the example pillar (not changed above), we see *both* owner and user listed for databases.
Furthermore, the formula also already uses

    {% if db.get('owner') %}
    - owner: {{ db.get('owner') }}
    {% endif %}


for the database object. Therefore, I contend that (at least!) changing the *dependency* to use "owner" vs "user" as above is correct. What I don't understand is why the pillar has a "user" attribute as well -- I don't see db.get('user') anywhere else.

dynjnelson:
If this would be easier to discuss this as a split PR I can do that!

dynjnelson:
OK. So *user* is for the *postgres user* that psql connects as (analagous to the 'runas' param which is for the UNIX user the command is executed as).  *Owner* is for the actual owner of the object.

Therefore, both user (which should default to 'postgres') and 'owner' are appropriate and necessary.
I would further argue that *both* user and owner (if specified) are required dependencies.

dynjnelson:
So it gets better (worse?)
Almost every part of the formula mis-uses owner/user/runas.
Example:

```
{% if db.extensions is defined %}
{% for ext, ext_args in db.extensions.items() %}
postgresql-ext-{{ ext }}-for-db-{{ name }}:
  postgres_extension.present:
    - name: {{ ext }}
    - user: {{ db.get('runas', postgres.user) }}
    - maintenance_db: {{ name }}
{% if ext_args is not none %}
{% for arg, value in ext_args.items() %}
    - {{ arg }}: {{ value }}
{% endfor %}
    - require:
      - service: postgresql-running
{% endif %}
{% endfor %}
{% endif %}
{% endif %}
{% endfor %}
```

That should have specified 'runas' not 'user'.

Would a PR that fixes these issues be a better choice?
I would argue that *every* object created by the formula needs three (optional!) items:

runas, user, and owner.

These would then mirror the actual module's usage of the values.

gravyboat:
Haha, I feel like we have discussed this confusion before in terms of the naming scheme for this formula. @babilen I'm down for a modification that adds support for the optional items and fixes things up. Do you have any concerns? It might break some backwards compatibility, but at some point a user is going to want these features.

dynjnelson:
@gravyboat @babilen 

A colleague and I are about to push a PR for the above changes.
We have an open question:  when it comes to "runas", should that be a per-substate config item or do people generally recommend a single global for runas defined in the postgres pillar which all components of a state share?

gravyboat:
I would say set a default which can then be overwritten in the pillar. Overwriting on a per-substate item, preferably it respects the configuration in the pillar which loops through the databases and respects the settings for each database item.

babilen:
@dynjnelson Thank you for your extensive analysis. I'd be more than happy to see this cleaned up.

dynjnelson:
Referenced in PR #110

iggy:
Chiming in a little late... Some of this code is quite old and runas is a new(ish) standard (it used to be just `user`). So that probably explains some of the confusion. Good job getting some of this cleaned up.

dynjnelson:
@babilen @iggy - I'd prefer if you made comments on #110 because that is where the bulk of this sort of change has taken place. Thanks!

PR #110, opened 2016-07-12
revamp cluster initialization, fix usage of user/owner/runas/db_user, add dependencies

dynjnelson:
@babilen, @gravyboat 

In no particular order, the following changes have been made:

1. the cluster initialization approach was revamped to both (hopefully) easier to use and more generic: now you specify in a "prepare_cluster" dictionary *four* things:

  1.  test -- a command to execute to determine if the cluster needs to be prepared/created
  2.  command -- the command to prepare the cluster
  3.  env -- a dictionary containing environment variables to set
  4.  user -- a system user to execute command with

  These four should be sufficient for pretty much every OS, easy to over-ride on a per-OS, per-OS version, per-OS subversion, and - of course - in the pillar.

2. The state was using 'user', 'runas', 'db_user', and 'owner' inconsistently and - in some cases - incorrectly. Often this worked because 'postgres' is both a common system user and postgresql user, however this is not true for all operating systems or environments. Furthermore, the way these are intended to be used is as such:

  * user - this is the *system* user that is used to execute psql. In modules, this is referred to as 'runas' frequently.
  * db_user - this is the *database* user that is used to authenticate and perform the operation
  * owner - this is the (future) database user/owner of the object created by db_user.

3. The approach taken in map.jinja did not appear to properly override defaults, os-specific configuration, or "codename" configuration, or configuration supplied in the pillar in a consistent fashion. This has been fixed along the same lines as the oft-referenced Apache formula and - in addition - we've supplied the capability of over-riding via the "osfinger" grain (which defaults to the empty dictionary here).

4. Inconsistent use of the "lookup" key was made consistent - the code does not make use of this key at all, however, this is an item that could use discussion.

5. The tablespace functionality gained the option to specify the OWNER of the new tablespace and properly quotes identifiers (vs. scalars).  The entire approach taken by salt in this manner, however, is suspect but also relegated to a future PR.

6. tablespace and extension configuration in the pillar now takes a different form: for example, for tablespaces, instead of taking a dictonary of tablespace name (the key) and directory (the value), now the data structure is a dictionary with the tablespace name as the key (this doesn't change) but a subdictionary of configuration for the value (which includes the directory). An analagous change was made for extensions.

7. Missing dependencies on objects were added:

  1. users created by a different user will require that the other user exists first.
  2. databases that are to be created by a given user will require that the given user exists first (and the same for database owner)
  3. schemas require the database they are to reside in to have been created first (ditto creating user and owning user)
  4. extensions require the database they are to reside in to have been created first (ditto creating user)

8. Arch no longer specifies arguments database creation that differ from every other OS (--locale en_US.UTF8 and -E UTF8), instead it uses the same mechanism as the other operating systems: namely, if you want to change the default encoding and locale, set it in the environment for prepare_cluster.  This allows for heterogeneus OS environments to maintain some increased level of consistency at the database layer.

We have tested on Ubuntu 14, Ubuntu 16, CentOS 7 and are in the process of building testing environments for Debian, Arch, and others.

To transition to the new data structure format, the following changes will need to take place:

1. if you had specified custom initdb commands, options, etc... transition those to the 'prepare_cluster' dictionary format.

2. if your pillar made use of the 'lookup' subkey, incorporate that sub dictionary into the main 'postgres' namespace. (See OPEN QUESTIONS below.)

3. If you were making use of extension or tablespace configuration, remember that each extension or tablespace now specifies a *dictionary* for the value rather than a string.

Example:

```
postgres:
  users:
    user-number-one:
      password: silliness
      superuser: True
    user-number-two:
      password: still-silly
  databases:
    db-number-one:
      owner: user-number-one
      schemas:
        extra_special_schema:
          owner: user-number-two
          user: user-number-one
      extensions:
        pgcrypto: {}
  tablespaces:
    fast-local:
      directory: /path/to/some/fast-local-storage
      owner: user-number-two
```

OPEN QUESTIONS:

1. the use or not-use of a 'lookup' subdictionary appears to be very inconsistently applied across salt formulas (formulae?), so it is not clear what the best practices here might be, especially given that all configuration is namespaced beneath "postgres".

2. Setting LC_ALL to C.UTF-8 doesn't seem to work everywhere - this probably needs to be revisted.

NOTE: We are pretty confident in the overall *structure* of these changes, modulo above noted issues, and welcome constructive feedback, etc!


pillar.example
```diff
@@ -1,17 +1,19 @@
 postgres:
   pg_hba.conf: salt://postgres/pg_hba.conf
-  commands:
-    initdb: service postgresql initdb
+  prepare_cluster:
+    user: root
+    command: service postgresql initdb
+    test: test -f /path/to/some/file
```
gravyboat: Should probably be a valid path just in case someone doesn't update this from a copy paste.

pillar.example
```diff
@@ -1,17 +1,19 @@
...
   use_upstream_repo: False
 
-  lookup:
-    pkg: 'postgresql-9.3'
-    pkg_client: 'postgresql-client-9.3'
-    pkgs_extra:
-      - postgresql-contrib
-      - postgresql-plpython
-    pg_hba: '/etc/postgresql/9.3/main/pg_hba.conf'
+  pkg: 'postgresql-9.3'
```
gravyboat: Do we need quotes wrapping this?

gravyboat:
Alright sorry for the delay in getting to this, been a busy week and I wanted to give this the time it deserves. Everything looks good to me with these changes. The benefits far outweigh the negatives associated with the backwards incompatible changes. I've added a few minor line notes that aren't a big deal. Regarding your questions:

```
the use or not-use of a 'lookup' subdictionary appears to be very inconsistently applied across salt formulas (formulae?), so it is not clear what the best practices here might be, especially given that all configuration is namespaced beneath "postgres".
```
Yeah it is inconsistent (the joys of open source), we've been using it more but some formulas either don't use it for the way they do merges, or simply haven't been updated to do so, or the user preferred to use a different name for the dict.

```
Setting LC_ALL to C.UTF-8 doesn't seem to work everywhere - this probably needs to be revisited.
```

I don't know if this is so much an issue on our end or a postgres problem. I couldn't find any issues on the main salt repo for this, but this doesn't mean there aren't any.

I'm giving these changes the :+1:, hopefully @babilen has some time to review and if he concurs we'll get this merged. If no one responds in a few days ping me and I'll merge it.

pillar.example
dynjnelson: No, but I didn't *change* that line per-se - I only de-dented it. I don't know what the policy is regarding (explicitly) quoting or not quoting strings.

pillar.example
dynjnelson: The problem is, of course, that the path will be different on every OS (which is why it's overridden later).

dynjnelson:
Regarding the issue with:
`Setting LC_ALL to C.UTF-8 doesn't seem to work everywhere - this probably needs to be revisited.`

we actually encountered that with CentOS-7 but not with Ubuntu. Before this PR is merged, I would want to understand more about what's not working there and fix it.

dynjnelson:
More information on the issue we hit with CentOS 7: CentOS-7 does not *provide* a locale for C.UTF-8. Some potentially relevant information located in the following links:

* https://sourceware.org/glibc/wiki/Proposals/C.UTF-8
* https://github.com/docker-library/python/issues/13

So I removed the *default* locale and encoding settings. This still allows administrators to set up whatever settings they want without providing a default that may or may not work (or be appropriate) for their particular environment.

Included in the most recent push are three changes to the *database* encoding, lc_type, and lc_collate settings:  previously, encoding *defaulted* to en_US.UTF-8 and now - like lc_type and lc_collate - it's only set if it is supplied explicitly.

tgoodaire:
@gravyboat: @dynjnelson and I wanted to ping you about this PR. Is there anything that we can do to  help get this merged, or are we waiting for someone to have a chance to review this further?

gravyboat:
I was waiting for other people to review it, but as no one did I'm merging it. Thanks for your work guys.

PR #115, opened 2016-09-02
Fix SLS rendering because of `pkgs_extra` Jinja variable isn't set by default

vutny:
Hi @gravyboat 

Currently, rendering of `init.sls` fails if there are no Pillars set, assuming we want to use only defaults:
```
[CRITICAL] Rendering SLS 'base:postgres' failed: Jinja variable 'dict object' has no attribute 'pkgs_extra'
```

This PR converts Jinja condition with loop in a compact `postgresql-extra-pkgs-installed` state which is  more clean and able to handle missing `pkgs_extra` attribute.

### DIFF
--- a/postgres/init.sls
+++ b/postgres/init.sls
@@ -47,13 +47,9 @@ postgresql-running:
     - require:
       - cmd: postgresql-cluster-prepared
 
-{% if postgres.pkgs_extra %}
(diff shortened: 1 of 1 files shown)


gravyboat:
Good catch, thanks.

PR #116, opened 2016-09-04
Set sane defaults when `use_upstream_repo` and `version` Pillars are set

vutny:
This PR allows to easy use of the formula when `use_upstream_repo` Pillar item is set.
PostgreSQL packages version `9.5` will be installed by default if `version` is not set to something else.

The main part is that it defines other items of `postgres` Pillar dict to work on Debian/Ubuntu/RHEL/CentOS/etc without a need to manually set package names, paths, custom commands and so on. Of course, any of such values could be overridden in later Pillar assignments.

Basically, it will be possible to get working PostgreSQL cluster installed from upstream repo right "out-of-the-box" by only setting such Pillar:
```yaml
postgres:
  use_upstream_repo: True
  # Optionally set version:
  #version: '9.5'
```

Pillar examples file updated accordingly with comments.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -1,12 +1,19 @@
 postgres:
-  pg_hba.conf: salt://postgres/pg_hba.conf
+  # Set True to configure upstream postgresql.org repository for YUM or APT
   use_upstream_repo: False
+  # Version to install from upstream repository
+  version: '9.3'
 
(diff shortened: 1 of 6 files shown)


gravyboat:
Good stuff, thanks.

PR #117, opened 2016-09-05
Introduce `osmajorreleasemap` for RedHat release specific commands

vutny:
Hi @gravyboat 

This PR removes legacy `osfingermap.yaml` file, which wasn't used (and I think it's useless), and adds `osmajorreleasemap.yaml` instead. The file allows to specify differences between major releases of different OS distributions. For now, only the differences between RHEL/CentOS/etc release versions have been covered. Unfortunately, it's impossible to use OS codenames here like for Debian/Ubuntu.

The motivation for this is that the same PostgreSQL build for each RedHat release has its own "default" method of preparing DB cluster, which honors system-wide settings, paths, environment variables and so on. Some of these specific `prepare_cluster` options were previously located in `osmap.yaml` file.

Side effects:
- SLS/YAML formatting cleanup
- logical state reordering
- use of modern `grains.filter_by` module features (available since Salt version `2015.5.0`)


gravyboat:
Thanks.

aboe76:
@vutny thanks for this fix, Tested it on Archlinux and it works great, no more issues.

PR #118, opened 2016-09-06
Better ACL handling in the `pg_hba.conf` file

vutny:
Hi @gravyboat 

This PR makes provisioning of `pg_hba.conf` file little bit better by:

* introducing default ACL entries in the `defaults.yaml` file instead of having them hardcoded in the template (easy to redefine in Pillar, without a need to rewrite the whole template)
* making pretty table formatting for all ACL entries (5 fixed width columns separated by spaces)
* setting more restrictive permissions (`600`)
* updating Pillar item format explanation in the `pillar.example` file


gravyboat:
Looks good, only issue I have is this is probably a backwards incompatible change. Going to leave it open for a bit to see if anyone has thoughts but it gets a thumbs up from me. If no on responds in a day or two ping me again and we'll merge.

javierbertoli:
I do agree with both of you, it's provides better management of `pg_hba.conf` but also is backward incompatible. Also, this way, I think it will break acls order when composing them with external pillars, which merge data from different files and expose them to salt.

What about adding two extra parameters, like `acls_prolog` and `acls_epilog` (which should, by default, match the default existing rules in pg_hba), and then we apply the result of adding `acls_prolog` + `acls` + `acls_epilog`? That way, while composing the `acls`, we can make sure the `prolog` and `epilog` are applied in the correct order?

vutny:
@gravyboat @javierbertoli Thanks for your feedback!

I would rather consider this is not a breaking change, but rather a fix to match an original design, if I got it right :smile: 

### TL;DR
You will get only the ACLs you configured explicitly in your Pillar. Nothing else.
No `acls` Pillar set? We have a common usable default values. The logic becomes very simple.

### Longer details

My key points:

1. The `acls` Pillar item format remains exactly the same, only the way of handling them was changed.
2. The `acls` list in the Pillar (external or SLS) is not going to be merged with default values, last configured Pillar source always wins. It's default Salt behavior, unless you've set `pillar_merge_lists: True`, but `default.yaml` entries will be discarded even in such case. That's how `grains.filter_by` function works.
3. Having explicit ACLs in order was the main motivation behind this change (now it is possible because of point No.2)

So, by default we will always have default hard-coded entry in the `pg_hba.conf`:
```
local   all             postgres                                peer
```
This is required to perform maintenance tasks.

Previously, we had other bunch of "static" lines at the bottom of the file:
```
# "local" is for Unix domain socket connections only
local   all             all                                     peer
# IPv4 local connections:
host    all             all             127.0.0.1/32            md5
# IPv6 local connections:
host    all             all             ::1/128                 md5
```

These entries are good "fall-back" configurations, which could be overridden with the ones taken from Pillar by Jinja code above in the template (previous version).

But! That's a source of big confusion, I think.

For example, I would like only to grant `my_user` account an ability to login on local system using socket connection with the password by setting such Pillar:
```yaml
acls:
  - ['local', 'all', 'my_user']
```

Also, I need to remember that any other system user which has corresponding user created in the PostgreSQL cluster will be able to login *without a password*, but I did not configured that *explicitly*.

Another case: I don't want to allow TCP connections at all, only local socket. And I was need to add another couple of entries to the ACL:
```yaml
acls:
  - ['local', 'all', 'my_user']
  - ['local', 'all', 'all', 'reject']
  - ['host', 'all', 'all', '0.0.0.0/0', 'reject']
  - ['host', 'all', 'all', '::/0', 'reject']
```

These are the steps I've used to follow to get my setup secure. And it wasn't so obvious without reading the template code.

It's easy to forget that such things are necessary, you need to look at the template to recall which are those "default" ACLs.  Now I need only one item in the list, **exact configuration I want to have**.
My changes allow to configure any combination of complex ACLs via Pillar without having a headache of considering influences of default predefined values or editing the template file each time.

I hope I explained it comprehensibly. Would be happy to get your comments.

gravyboat:
Thanks for the detailed explanation @vutny, I'm going to merge this for now. Worst case we can always back it out.

vutny:
@gravyboat Sure, no worries. And I have more improvements to submit :smile:

PR #119, opened 2016-09-09
Map Pillar dicts for PostgreSQL entities to the Salt state arguments

vutny:
Hi @gravyboat 

This PR introduces a breaking change in Pillar layout for such PostgreSQL entities as `users`, `tablespaces`, `databases`, `schemas` and `extensions`.
Basically it gets the above dictionaries and converts underlying structure to the `postgres` state module syntax to render in the resulting SLS file. This is a pretty straight forward approach to support all arguments provided by particular state without a need to put `if` condition for each of them. Also it becomes natural to refer Salt documentation for the complete list and description of supported arguments.
Added explanation about this in `pillar.example` file.

List of backwards-incompatible changes:
* Previously, the `runas` argument in the Pillar was mapped to the `user` arg, and the `user` Pillar arg was mapped to the `db_user` arg in the resulting state. This is totally confusing, undocumented and hard to figure out from SLS code. Now the `user` is really the `user` and `db_user` == `db_user`.
* All the entities now support the `ensure` argument (including tablespaces).
* Secure permissions (`700`) is applied to a tablespace directory.
* The `schemas` becomes a standalone dict at the same level as `databases` or `users`, its name and arguments are mapped to the `postgres_schema` state.
* PostgreSQL extensions cannot be created *in database*, they are installed in some schema or globally. So, the `extensions` dict is also separated like the `schemas` dict, and, as expected, the arguments are mapped to the `postgres_extension` state.

Overall, these changes just make things work as expected, without a need to edit SLS if there would be any new args in the `postgres` states, the SLS becomes more compact and readable, easier to debug.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -31,11 +31,39 @@ postgres:
   # METHOD could be omitted, 'md5' will be appended by default.
   acls:
     - ['local', 'db1', 'localUser']
-    - ['host', 'db2', 'remoteUser', '123.123.0.0/24']
+    - ['host', 'db2', 'remoteUser', '192.168.33.0/24']
 
   # PostgreSQL service name
   service: postgresql
 
+  # Create/remove users, tablespaces, databases, schema and extensions.
+  # Each of these dictionaries contains PostgreSQL entities which
+  # mapped to the `postgres_*` Salt states with arguments. See the Salt
+  # documentaion to get all supported argument for a particular state.
+  #
+  # Format is the following:
+  #
+  #<users|tablespaces|databases|schemas|extensions>:
+  #  NAME:
+  #    ensure: <present|absent>  # 'present' is the default
+  #    ARGUMENT: VALUE
+  #    ...
+  #
+  # where 'NAME' is the state name, 'ARGUMENT' is the kwarg name, and
+  # 'VALUE' is kwarg value.
+  #
+  # For example, the Pillar:
+  #
+  #users:
+  #  testUser:
+  #    password: test
+  #
+  # will render such state:
+  #
+  #postgres_user-testUser:
+  #  postgres_user.present:
+  #    - name: testUser
+  #    - password: test
   users:
     localUser:
       ensure: present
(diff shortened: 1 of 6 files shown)


javierbertoli:
Hi @vutny! IMHO, awesome refactoring! :smile:

I'd just modify the `pillar.example` to make tests pass, as current values and/or format are not completely functional. Don't know if you want to write them or, else, I can try to (on this same branch), probably later today/tomorrow?

vutny:
@javierbertoli Thanks.

Hm... Which kind of test failures have you seen? They should work as before.
Could you please share the output you got?

javierbertoli:
Sure, will run them later again and post them.

From the top of my head, I remember that when I run kitchen-salt to test this branch, I found:

* Debian's default is now 9.4, not 9.3 (there's no 9.3 available anymore)
* `db2` creation fails, as the `user` is `remoteUser`, which does not exist in the system (`user` refers to system user, not db_user). Didn't even finish runing the serverspec tests or the other distros.

 I run this branch with a modified version of my production pillar data and it worked OK, so that's why I bet it's mostly a matter of tweaking the pillar.example data to a valid working set :)

vutny:
@javierbertoli Understood, thanks a lot! I missed that stuff after merging my from my own playground.
I'll fix it soon.

vutny:
@javierbertoli I've fixed example Pillar data and the tests. Please take a look.

Actually, I've also hit a minor bug in Salt's `postgres` module -- it returns `True`, but no changes. Will need to submit a fix there too, but the formula works and usable in any case.

Regarding 9.4 on Debian: indeed, that's already covered for `jessie` in `codenamemap.yaml` file. But the tests are using Ubuntu 14.04 image, which is still on 9.3 version from their repos.

javierbertoli:
@gravyboat, @vutny I think this is OK to be merged. (@gravyboat I don't know if I'm allowed to merge it, or there's an agreement on who's on charge of doing it?)

Also, wouldn't at least a `VERSION` file, using Semantic Versioning, be useful? That'd be a good way to know when backward-incompatible changes are merged, don't you agree?

gravyboat:
@javierbertoli You can merge it if you have the perms to do so, I'm always happy to share the load of reviewing all of these with other people.

PR #120, opened 2016-09-13
Make setting `use_upstream_repo` install PostgreSQL 9.5 by default

vutny:
Hi @gravyboat 

This PR is a follow-up for the #116 and it makes that feature actually work for Debian/Ubuntu as well.
Previously, you was required to specify `postgres:version` Pillar item for the `upstream.sls` to be rendered without an error, but RHEL systems family were OK.

Now the behavior for both platforms becomes consistent and all version dependent arguments for `pkgrepo.managed` state are expressed as `postgres` dictionary items set in `*map.yaml` files.

### DIFF
--- a/postgres/codenamemap.yaml
+++ b/postgres/codenamemap.yaml
@@ -5,7 +5,8 @@
   {% endif %}
 
 {{ name }}:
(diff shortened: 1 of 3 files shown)


gravyboat:
Good stuff, thanks!

PR #122, opened 2016-09-19
Debian: make codenamemap even if `lsb-release` pkg not installed

vutny:
Hi @gravyboat

### The Issue

Salt relies on `lsb-release` package to provide correct distribution codename in the `oscodename` Grain on Debian GNU/Linux systems.
If that package is not installed (e.g. you have minimal system or a container), Salt provides different string as  `oscodename`, this is causing the SLS compilation errors.

### The Fix

It's virtually impossible to install needed package and refresh Grain data during construction of `postgres` lookup dictionary, even using `salt` calls in Jinja. But this is a required step for compiling SLS files.

Instead, optional parameter was added to the macros which generates lookup dictionary map for Debian distros.
Now the both cases are covered: with and without `lsb-release` installed, and the formula works as expected.

### DIFF
--- a/postgres/codenamemap.yaml
+++ b/postgres/codenamemap.yaml
@@ -1,12 +1,25 @@
-{% macro ubuntu_block(name, version) %}
+### Set parameters based on PostgreSQL version supplied with particular distro
 
-  {% if salt['pillar.get']('postgres:use_upstream_repo', False) %}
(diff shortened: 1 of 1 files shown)


gravyboat:
Good stuff, thanks @vutny.

PR #124, opened 2016-09-22
Support only including postgres.client

tobio:
In the current state, we must include `postgres` to satisfy the `require_in` statement in the upstream repo states. This moves the requisites to the `postgresql-installed` state allowing nodes to only include the `postgres.client` file.


gravyboat:
Good stuff, thanks.

PR #126, opened 2016-09-26
RedHat: make client binaries available in `$PATH`

vutny:
PostgreSQL client RPM package, which comes from upstream repository, puts all binaries in a separate directory: `/usr/pgsql-9.X/bin`. Obviously, such dir is not included in the `$PATH` by default, and any tools (e.g. Salt modules) which rely on `psql` command just can't found that binary file in place.
Also, as a system operator, I assume that all commands provided by client package will be accessible immediately after installation without specifying full location.

This PR adds `alternative.install` state for each PostgreSQL client executable found to create corresponding symlink in `/usr/bin`. Updating the `$PATH` globally would be a bad choice, because it requires reloading system profile or reestablishing current login sessions, without considering that the `$PATH` env variable possibly could be overridden elsewhere. 

Additional improvement: force refreshing of repo metadata only if `postgres.use_upstream_repo = True`, but otherwise let the Salt decide what to do depending on state run (instead of disabling refresh at all).


vutny:
Ping @javierbertoli

postgres/defaults.yaml
```diff
@@ -39,3 +39,30 @@ postgres:
...
+    - pg_dump
+    - pg_dumpall
+    - pg_isready
+    - pg_receivexlog
+    - pg_restore
+    - pg_rewind
+    - pg_test_fsync
+    - pg_test_timing
+    - pg_upgrade
+    - pg_xlogdump
```
javierbertoli: These binaries `pg_rewind
pg_test_fsync
pg_test_timing
pg_upgrade
pg_xlogdump
` do not exist in my Debian's installation, so I don't think this list belongs into the 'default.yaml' file, it seems to be specific to the upstream repo.

vutny: @javierbertoli Thanks for review. Indeed, because this addition is totally RedHat specific (for upstream packages), its definitely better to set `client_bins` in the `osmap.yaml`. I'll do that.

But, the `bin_dir` item already belongs to the `RedHat` section in the `osmap` after the "`{% if repo.use_upstream_repo %}`" switch. Also, I've added `onlyif` requisite to the `alternatives.install` state in the `client.sls`. So, the cases with other platforms and even some missing binaries are covered as well.

vutny:
@javierbertoli I've pushed changes according to the comments above. Please take a look.

javierbertoli:
Awesome @vutny, thanks! 😃

PR #127, opened 2016-10-05
Manage PostgreSQL entities with `client.sls` states

vutny:
Hi.

This PR moves away and splits states from `init.sls` to:

* `server.sls`: installs PostgreSQL server package and prepares the DB cluster
* `client.sls`: additionally responsible for creating DB objects (users, tablespaces, databases, etc)

Motivation behind this change is to be able to manage some PostgreSQL server installation which is out of Salt control ([AWS RDS](https://aws.amazon.com/rds/postgresql/) could be a good example) with `client.sls`. It will still allow to manage PostgreSQL DB entities with Salt (even with build-in modules), but eliminates the requirement to have the server installed locally. But client binaries are required to achieve that, of course.

Applying the `init.sls` (`postgres`) states will give exactly the same result as it was before.


gravyboat:
Good stuff, thanks.

javierbertoli:
@vutny I was about to comment that DB objects creation does not belong to a `client` state to me (as we discussed before, it's not 'intuitive'), but @gravyboat merged it already 😃 

I wonder if separating states, like:

* postgres/database.sls
* postgres/users.sls
* postgres/schema.sls
* postgres/extension.sls

would make it easier to manage each of these DB objects without issues, and keeping again states names intuitive?

And as you did, use them all in the `postgres/init.sls` to maintain compatibility with current behaviour?

gravyboat:
@javierbertoli We can always back it out or prep a new PR, let me know what you guys want to do.

vutny:
@gravyboat Thanks for the quick reaction :smile: 

@javierbertoli Thank you for the feedback!
I think that putting DB objects provisioning in client state is being logical because of following reasons:

* All these entities are strictly depending on the PostgreSQL client binaries which Salt uses to get the job done. The states are just unavailable if Salt is unable to find `psql`, for example, in the right place.
* You could set all parameters configuration only via Pillar interfaces, explicitly targeting which "client" Minion should provision some external DB instance. Having also to figure out and make additional changes to the `top.sls` for states is rather non-intuitive and unnecessary.
* We have such behavior already documented in README :tada: You just need to apply `client.sls` to get things done *instead of* calling the `init.sls` and waiting for the full-blown PostgreSQL to be installed for the sake of just creating a user in *remote DB*.
* The `init.sls` still works as it worked before, default behavior hasn't been changed. Both server and client will be installed on the same Minion. But we additionally make sure that client is set up correctly and available for Salt.

I agree that for getting more modularity and readability we could move DB objects states to the `objects.sls` file which will depend on the `client.sls`. But there's no value in scattering the states too much over many SLS files.

javierbertoli:
Hey, @vutny!

This is, of course, a discussion more focused on design than anything else, so take this into consideration when reading my answer. I don't consider your points 'invalid' or wrong, nor thing that this PR is wrong either. It's just that I consider some aspects of the design differently 😄 

I like modules to *reflect the way the application they configure behaves and how the application parts interact*, because I find that more *intuitive*  to the newcomer to the module. She may not know the module, but already has some knowledge of the application she's going to configure, and her mindset will probably try to follow that knowlege when trying to understand the module. 

(I didn't realize that the postgres states in salt use the `psql` client. I thought they use some `libpq` wrapper like [psycopg](http://initd.org/psycopg/) or similar. So I see the reasoning in your 1st. point.) 

Still, I expect a `client` state to just install the `client` app. If, at the moment, *the presence of the client app is  a requirement to configure PG objects*, that's consequence of the way Saltstack designed its PG management states and modules. And therefore, "wherever you configure and manage the PG objects", you'll need to include the `client` binary to satisfy this need. But not the other way round.

In my opinion, the way you distributed things is logical and intuitive from **the Saltstack point of view**, but not fom the **Postgresql** point of view. With the reasoning you followed, tomorow, when Saltstack postgres module moves to psycopg, we'd have to move all the PG objects configuration and management to a state named `pypgclient.sls`?

Lastly, I agree with you, and neither do I like the idea of scattering configurations in so many parts. That's why I started with "I wonder" in my previous example: it was a quick list to illustrate my point, with no much thought. Still, I think that something like:

* `server.sls`: installs and initializes the server
* `client.sls`: install the client binary
* `config.sls`: manage the server (and client) configurations (if you don't want to add this state, you can add the configuration resources in the corresponding `server` and `client` states)
* `manage.sls`? (or other meaningful name you can think of): manages users, databases, schemas, extensions and other PG objects we can add. *Requires* the `client.sls` because of Salt's design.
* init.sls: include all of the above, to keep compatibility.

vutny:
Hi @javierbertoli 

Your comments seem very reasonable, I totally agree. Let me prepare another PR to address this properly.

Only one thing I would like to avoid: making separate `config.sls` file. I've seen that in other formulas and it doesn't look solid. Most often (like with PostgreSQL) configuration files depend on package(s) which makes necessary directory tree and creates users, but also introduced services rely on those configs. Basically, purely configuration states just can't be stand alone and reusable.

I'm even thinking about making `manage.sls` decoupled from `client.sls`... If you provision PostgreSQL binaries without Salt (by producing VM/container images for example), there is no need to automatically include client states as dependencies. It's rather better to check if Salt is able to load `postgres_*` modules and act accordingly (include `client.sls` or not).

What do you think?

PR #128, opened 2016-10-09
Fix missing spaces in long pg_hba.conf lines

Laogeodritt:
In generated pg_hba.conf, fix missing spaces between arguments if an argument is longer than the column width allotted to that argument in the line.

An example pillar excerpt:

    postgres:
        # ...
        acls:
            - ['local', 'database1', 'samehost', 'user']
            - ['local', 'database1,database2,database3', 'user', 'samehost', 'md5']

Before patch, outputs are:

    local    database1       user            samehost                md5
    local   database1,database2,database3user            samehost                md5

After patch:

    local    database1       user            samehost                md5
    local   database1,database2,database3 user            samehost                md5

Second line is less readable since the columns aren't strictly adhered to anymore, but it is correctly read by postgresql at least.

### DIFF
--- a/postgres/templates/pg_hba.conf.j2
+++ b/postgres/templates/pg_hba.conf.j2
@@ -38,5 +38,5 @@ local   all             postgres                                peer
     {%- endif %}
 
   {%- endif %}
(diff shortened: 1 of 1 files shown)


javierbertoli:
Great catch! Thanks!

PR #129, opened 2016-10-20
[REFACTORING] Add `postgres.manage` state to provision DB objects

vutny:
This is a follow-up for the PR #127 which introduces `postgres.manage` state according to the discussion we have took there.

So, client binaries would be provisioned only when Salt is unable to call `postgres` exec module functions during the (high)state run. This makes "standalone" provisioning of PostgreSQL DB much faster on systems or appliances were server/client binaries are already available.


gravyboat:
Good stuff, thanks.

PR #130, opened 2016-11-04
Remove outdated modules for managing PostgreSQL tablespaces

vutny:
This PR removes previously imported state and execution module for managing PostgreSQL tablespaces.

The reasons for this are that PG tablespaces support is available in Salt since release `2015.8` (oldest supported branch) and some fixes were made there (saltstack/salt#36464) which should be included to point-releases `2015.8.13` and `2016.3.4`.

So, the formula does not need to ship old modules with it.


javierbertoli:
Thanks @vutny !

PR #132, opened 2016-12-01
Workaround a bug in saltstack/salt#37935 for release v2016.11.0

vutny:
This PR adds a workaround for Salt issue https://github.com/saltstack/salt/issues/37935 which was introduced in release `2016.11.0`. It seems to be fixed now by PR saltstack/salt#37993, and the fix should be available in the future point release.

Till that moment we still have the latest stable Salt packages available at https://repo.saltstack.com which do have mentioned bug. It affects all PostgreSQL versions for RHEL systems installed from the upstream repository.

Proposed workaround is to make sure that the PostgreSQL server binaries also available in the system `$PATH`, as it was done with client executables earlier. It also may be useful for some people managing parts of the installation manually.
The `postgresql.manage` state will include a dependency for the `postgres.server` state (which installs binaries) only for particular Salt version `2016.11.0` on RHEL platforms.

Ping @gravyboat

### DIFF
--- a/postgres/manage.sls
+++ b/postgres/manage.sls
@@ -8,6 +8,10 @@
 
 include:
   - postgres.client
+  {%- if 'server_bins' in postgres and grains['saltversion'] == '2016.11.0' %}
+  # FIXME: Salt v2016.11.0 bug https://github.com/saltstack/salt/issues/37935
+  - postgres.server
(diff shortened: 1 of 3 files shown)


gravyboat:
Good stuff, thanks for addressing this!

PR #135, opened 2016-12-19
Fix and set useful backup extension for postgresql.conf file

vutny:
This PR provides better and more flexible configuration for making a backup of `postgresql.conf` file.

Some earlier changes caused that the name of resulting backup file appeared to be `postgresql.confTrue` when using defaults, which is not pretty at all. That's fixed.

Also there is an example of how to set useful custom backup file extension with timestamp suffix, it should come in handy when you really will need to restore something.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -14,9 +14,10 @@ postgres:
     - postgresql-contrib
     - postgresql-plpython
 
-  # This section will append your configuration to postgresql.conf.
+  # Append the lines under this item to your postgresql.conf file.
+  # Pay attention to indent exactly with 4 spaces for all lines.
   postgresconf: |
-    listen_addresses = 'localhost,*'
+    listen_addresses = '*'  # listen on all interfaces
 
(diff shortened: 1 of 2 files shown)


vutny:
I've also found that the current example entry of `listen_addresses` option for the `postgresql.conf` file is meaningless: the PostgreSQL server is listening on `localhost` only in that case, ignoring the special `*` sign after comma. So, corrected it as well.

vutny:
@gravyboat @javierbertoli I hope you enjoyed your winter and Christmas holidays if you had some of it :smile: 

Could anyone please take a look at this PR and merge if it's OK? Thank you.

gravyboat:
Looks fine to me, as a heads up @vutny I've stepped down from maintaining the formulas repos so I won't be getting notifications unless I'm specifically mentioned, probably better to tag someone on the Salt team going forward.

javierbertoli:
Hi @vutny!, Thanks for the PR. Merging.

gravyboat:
@vutny You're welcome!

PR #137, opened 2017-01-12
Skip configuring ACLs in `pg_hba.conf` if not provided

vutny:
Hi @javierbertoli 

This PR adds possibility to skip ACL configuration in `pg_hba.conf` (use distribution or already existing file) when the `postgres:acls` Pillar contains empty value such as string(`''`), arrary (`[]`) or `null` (equivalent of not giving any value). In any case the state will ensure that the file exists and has correct permissions.

This becomes very handy if you're fine with packaged config, or would like to convert existing PostgreSQL installation to be supported by Salt, or ACLs need to be managed manually with other tools.

Could you please review this one? Thank you!

P.S. Some typos has been fixed.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -41,6 +41,9 @@ postgres:
   #
   # The uppercase items must be replaced by actual values.
   # METHOD could be omitted, 'md5' will be appended by default.
+  #
+  # If ``acls`` item value is empty ('', [], null), then the contents of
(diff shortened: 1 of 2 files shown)


babilen:
Great, that will come in handy.

vutny:
Thanks for the quick merge @babilen !

PR #138, opened 2017-01-13
Make local backup of `pg_hba.conf` file before modification

vutny:
This PR enables local backup creation of modified `pg_hba.conf` file by default.
The `config_backup` Pillar setting has been introduced to trigger backup for all modified PostgreSQL server configuration files (`postgresconf` and `pg_hba.conf`).

It's very basic and useful feature, I think. @babilen or @javierbertoli , could you please review this? Thank you.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -19,12 +19,6 @@ postgres:
   postgresconf: |
     listen_addresses = '*'  # listen on all interfaces
 
-  # Backup extension for postgresql.conf file, defaults to ``.bak``.
-  # Set to False to stop creation of backup on postgresql.conf changes.
-  {%- if 'status.time' in salt.keys() %}
-  postgresconf_backup: ".backup@{{ salt['status.time']('%y-%m-%d_%H:%M:%S') }}"
-  {%- endif %}
-
   # Path to the `pg_hba.conf` file Jinja template on Salt Fileserver
   pg_hba.conf: salt://postgres/templates/pg_hba.conf.j2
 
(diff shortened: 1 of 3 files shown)


EvaSDK:
Any specific reason to choose the local copy of pg_hba to `.bak` instead of using the `file.managed` backup system like the `postgresql-conf` state ?

vutny:
Yeah, there is a reason. Good question, thanks! :+1: 

The `backup` parameter works completely different in `file.blockreplace` and `file.managed` states.
The approach from `file.blockreplace` is naive: make the backup at the same directory and append provided backup extension to the file name.

The `file.managed` state processes `backup` argument in more complex way, see: https://docs.saltstack.com/en/latest/ref/states/backup_mode.html
This is more advanced, but not always convenient. Especially if someone else manages PostgreSQL with other tools than Salt. Also, you can configure that in a Minion config.

That's why I "emulated" the behavior of `backup` to be similar to `postgresql-conf` (`file.blockreplace`), but just for `pg_hba.conf` (`file.managed`) using separate `file.copy` state.

EvaSDK:
Ok, I thought that `file.blockreplace` and `file.managed` were behaving the same so 👍 for consistency.

pillar.example
```diff
@@ -17,13 +17,7 @@ postgres:
...
   # Pay attention to indent exactly with 4 spaces for all lines.
   postgresconf: |
-    listen_addresses = '*'  # listen on all interfaces
-
-  # Backup extension for postgresql.conf file, defaults to ``.bak``.
-  # Set to False to stop creation of backup on postgresql.conf changes.
-  {%- if 'status.time' in salt.keys() %}
-  postgresconf_backup: ".backup@{{ salt['status.time']('%y-%m-%d_%H:%M:%S') }}"
-  {%- endif %}
+      listen_addresses = '*'  # listen on all interfaces
```
YetAnotherMinion: Looks like there was an indentation change with list_addresses?

myoung34: Likely not on purpose, should be fixed before merge

vutny: Whoops, I guess I switched context too often and started to mix things :)
Indeed, YAML (and SLS) literal style scalars should be indented exactly with two spaces. Fixed, thanks!

vutny: And since it is just a multi-line string which is directly used in the `file.blockreplace` `context` parameter in the same way (literal style), there should be no `-` list indicator.

vutny:
@javierbertoli, could you please review this? Many thanks.

javierbertoli:
Looks good to me. Thanks @vutny and @EvaSDK!

PR #139, opened 2017-01-24
add example on how to enable extension on a db

ek9:
add example on how to enable extension on a db

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -131,6 +131,14 @@ postgres:
       lc_ctype: 'en_US.UTF-8'
       lc_collate: 'en_US.UTF-8'
       tablespace: 'my_space'
+      # set custom schema
+      schemas:
+        public:
+          owner: 'localUser'
+      # enable per-db extension
+      extensions:
+        uuid-ossp:
+          schema: 'public'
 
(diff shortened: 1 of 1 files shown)


javierbertoli:
Thanks @ek9!

YetAnotherMinion:
Just a note about this per db extensions. I see warning that `schemas` and `extensions` will not work in Nitrogen version of SaltStack. I made a naive attempt to squash this error by moving the per database extensions and schemas under a `context` key, but that just changed the warning to  `context` being an invalid keyword for `postgres_database.present`. I don't know enough about salt yet to know exactly where the `postgres_database_present` expects to find the `context` parameter.

```
 Warnings: 'extensions' and 'schemas' are invalid keyword arguments for
              'postgres_database.present'. If you were trying to pass additional
              data to be used in a template context, please populate 'context'
              with 'key: value' pairs. Your approach will work until Salt
              Nitrogen is out. Please update your state files.
```

ek9:
@YetAnotherMinion perhaps you could open a separate issue for this (call it something related to `Nitrogen extension and schema` support) . I could have a look later today.

PR #155, opened 2017-08-14
Add support for using the same dict key, but different 'name's

Lucianovici:
Hi,

I have a use case where in some other jinja templates I reference these pillar data. So I need to have a steady key and the `name` should be different for each environment.

Do you think this PR makes sense?

e.g. postgres pillar for dev env:

```
  users:
    devices_user:
      name: dev_devices_user
      ensure: present
      password: 'password'

  databases:
    devices:
      name: dev_devices
      owner: 'dev_devices_user'
      template: 'template0'
```

e.g. postgres pillar for test env:

```
  users:
    devices_user:
      name: test_devices_user
      ensure: present
      password: 'password'

  databases:
    devices:
      name: test_devices
      owner: 'test_devices_user'
      template: 'template0'
```
e.g. a Jinja template that configures my Django databases for different envs:

```
DATABASES:
  default:
    ENGINE: 'django.db.backends.postgresql'
    NAME: '{{ salt.pillar.get('postgres:databases:devices:name') }}'
    USER: '{{ salt.pillar.get('postgres:users:devices_user:name') }}'
    PASSWORD: '{{ salt.pillar.get('postgres:users:devices_user:password') }}'
```

### DIFF
--- a/postgres/manage.sls
+++ b/postgres/manage.sls
@@ -25,7 +25,7 @@ postgres-reload-modules:
 
 {%- for key, user in postgres.users|dictsort() %}
 
-{{ format_state(user.name or key, 'postgres_user', user) }}
+{{ format_state(user.get('name') or key, 'postgres_user', user) }}
     - require:
       - test: postgres-reload-modules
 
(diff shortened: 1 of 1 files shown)


johnkeates:
Wouldn't this break existing setups? Also, aren't envs better used using Salt's own env support?

Lucianovici:
Well, I'm using the Salt's own env support. I don't really get it. I'm a bit new to this salt world :) My yamjam Django database config sits in a template and it's used in `states` to ensure my supervised gunicorn server will have the correct database config for each salt env.

Now It won't break the existing setup since there is `entity.get('name') or key` so if there is no `name` given, it will use the `key`.

Is there a better way of accomplishing what I'm looking for?

aboe76:
#223 is merged closing this one.

PR #169, opened 2017-09-25
Suse: Introduce support for upstream repo

noelmcloughlin:
This PR is an enhancement to allow the [upstream repo](https://download.postgresql.org/pub/repos/yum/) to be used for SLES and OpenSUSE.

The osmap.yaml was updated introducing the feature. Design is  aligned with Redhat Stanza for convenience and continuity.  These log files demonstrate working "feature" on OpenSuse LEAP 42 as postgresql 9.5 software is retrieved from upstream-

1. [postgres95_nogpgcheck_run1.log.txt](https://github.com/saltstack-formulas/postgres-formula/files/1327882/postgres95_nogpgcheck_run1.log.txt)
2. [postgres95_nogpgcheck_run2.log.txt](https://github.com/saltstack-formulas/postgres-formula/files/1327883/postgres95_nogpgcheck_run2.log.txt)
3. [postgres95_nogpgcheck_run3.log.txt](https://github.com/saltstack-formulas/postgres-formula/files/1327884/postgres95_nogpgcheck_run3.log.txt)
 
**Limitation**
GPGCHECK is disabled (for now): See https://redmine.postgresql.org/issues/2732 

**Followup**
1. Fix gpgcheck on Suse - Raise ISSUE.

**Nice to Have**
The pillar.example was improved in the following manner-
- Required parameters are prominent.  For example the value of **service** differs depending on whether linux distribution or upstream repo is used.
- Unnecessary parameters (pkg, pkg_client, pkgs_extra) were removed to fix confusion.

Thanks

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -1,17 +1,18 @@
 postgres:
-  ## USE DISTRO REPO
+  # Set True to configure upstream postgresql.org repository for YUM or APT
   use_upstream_repo: False
-  service: postgresql
+  # Version to install from upstream repository
+  version: '9.3'
 
-  ## OR UPSTREAM REPO (Suse/Redhat)
-  # use_upstream_repo: True
-  # version: '9.5'
-  # service: postgresql-9.5
+  # These are Debian/Ubuntu specific package names
+  pkg: 'postgresql-9.3'
+  pkg_client: 'postgresql-client-9.3'
 
-  ## AND OVERRIDES
   # Additional packages to install with PostgreSQL server,
   # this should be in a list format
-  pkgs_extra: []
+  pkgs_extra:
+    - postgresql-contrib
+    - postgresql-plpython
 
   # Append the lines under this item to your postgresql.conf file.
   # Pay attention to indent exactly with 4 spaces for all lines.
@@ -46,6 +47,9 @@ postgres:
   {%- if salt['status.time']|default(none) is callable %}
   config_backup: ".backup@{{ salt['status.time']('%y-%m-%d_%H:%M:%S') }}"
   {%- endif %}
+
+  # PostgreSQL service name
+  service: postgresql
 
   {%- if grains['init'] == 'unknown' %}
 
--- a/pillar.example
+++ b/pillar.example
@@ -1,18 +1,17 @@
 postgres:
-  # Set True to configure upstream postgresql.org repository for YUM or APT
+  ## USE DISTRO REPO
   use_upstream_repo: False
-  # Version to install from upstream repository
-  version: '9.3'
+  service: postgresql
 
-  # These are Debian/Ubuntu specific package names
-  pkg: 'postgresql-9.3'
-  pkg_client: 'postgresql-client-9.3'
+  ## OR UPSTREAM REPO (Suse/Redhat)
+  # use_upstream_repo: True
+  # version: '9.3'
+  # service: postgresql-9.3
 
+  ## AND OVERRIDES
   # Additional packages to install with PostgreSQL server,
   # this should be in a list format
-  pkgs_extra:
-    - postgresql-contrib
-    - postgresql-plpython
+  pkgs_extra: []
 
   # Append the lines under this item to your postgresql.conf file.
   # Pay attention to indent exactly with 4 spaces for all lines.
@@ -47,9 +46,6 @@ postgres:
   {%- if salt['status.time']|default(none) is callable %}
   config_backup: ".backup@{{ salt['status.time']('%y-%m-%d_%H:%M:%S') }}"
   {%- endif %}
-
-  # PostgreSQL service name
-  service: postgresql
 
   {%- if grains['init'] == 'unknown' %}
 
--- a/pillar.example
+++ b/pillar.example
@@ -5,8 +5,8 @@ postgres:
 
   ## OR UPSTREAM REPO (Suse/Redhat)
   # use_upstream_repo: True
-  # version: '9.3'
-  # service: postgresql-9.3
+  # version: '9.5'
+  # service: postgresql-9.5
 
   ## AND OVERRIDES
   # Additional packages to install with PostgreSQL server,
--- a/postgres/osmap.yaml
+++ b/postgres/osmap.yaml
@@ -94,23 +94,24 @@ Suse:
   pkg_repo:
     name: pgdg-sles-{{ release }}
     humanname: PostgreSQL {{ repo.version }} $releasever - $basearch
-    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/suse/sles-$releasever-$basearch'
-    gpgcheck: 1
-    gpgkey: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/suse/sles-$releasever-$basearch/repodata/repomd.xml'
+    #Suse repos are suffixed 'suse/sles-12-$basearch' by upstream community convention.
+    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/suse/sles-12-$basearch'
+    gpgkey:  'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/suse/sles-12-$basearch/repodata/repomd.xml'
+    #future: fix suse "4-Signatures public key is not available" issue and enable gpgcheck.
+    gpgcheck: 0
 
 {% if repo.use_upstream_repo %}
 
-  {% set data_dir = '/var/lib/pgsql/' ~ repo.version ~ '/data' %}
+  {% set lib_dir = '/var/lib/pgsql/' ~ repo.version ~ '/data' %}
 
   pkg: postgresql{{ release }}-server
   pkg_client: postgresql{{ release }}-server
-  conf_dir: /var/lib/pgsql/{{ repo.version }}/data
-  bin_dir: /usr/pgsql-{{ repo.version }}/bin
+  conf_dir: {{ lib_dir }}
   service: postgresql-{{ repo.version }}
 
   prepare_cluster:
-    command: initdb --pgdata='{{ data_dir }}'
-    test: test -f '{{ data_dir }}/PG_VERSION'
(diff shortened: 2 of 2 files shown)


postgres/osmap.yaml
```diff
@@ -91,9 +91,70 @@ RedHat:
...
+  pkg_client: postgresql{{ release }}-server
+  conf_dir: {{ lib_dir }}
+  service: postgresql-{{ repo.version }}
+
+  prepare_cluster:
+    command: initdb --pgdata='{{ lib_dir }}'
+    test: test -f '{{ lib_dir }}/PG_VERSION'
+
+  # Directory containing PostgreSQL client executables
+  bin_dir: /usr/pgsql-{{ repo.version }}/bin
```
vutny: Why to blindly copy-paste all the stuff? It seems that SUSE packages install all (?) needed alternatives itself during post-installation phase.

postgres/osmap.yaml
```diff
@@ -91,9 +91,70 @@ RedHat:
...
+    - postgres
+    - postgresql{{ release }}-check-db-dir
+    - postgresql{{ release }}-setup
+    - postmaster
+{% else %}
+
   pkg: postgresql-server
-  pkg_client: postgresql
-  pkg_libpq_dev: postgresql
+  pkg_client: postgresql-server
```
vutny: Again, are you sure the client and the server package are the same in packages provided by SUSE distribution?

pillar.example
```diff
@@ -1,18 +1,17 @@
...
+  # version: '9.5'
+  # service: postgresql-9.5
 
+  ## AND OVERRIDES
   # Additional packages to install with PostgreSQL server,
   # this should be in a list format
-  pkgs_extra:
-    - postgresql-contrib
-    - postgresql-plpython
+  pkgs_extra: []
```
vutny: I'm aware that hard-coding platform-specific packages here was bad idea. But this possibly may affect later installation of the `uuid_ossp` extension defined below. Have you tried to run test kitchen with provided config?

noelmcloughlin: No, I've tested on OpenSUSE LEAP 42.

postgres/osmap.yaml
noelmcloughlin: Yes, I'm sure.

postgres/osmap.yaml
noelmcloughlin: The PR description explains why ... "Design is aligned with Redhat Stanza for convenience and continuity".  It keeps the PR simple.

>It seems that SUSE packages install all (?) needed alternatives itself during post-installation phase.
 
The upstream repo is common to SusE/RedHat/Fedora.  The alternatives handling can therefore be improved for these distributions but not in this PR.

postgres/osmap.yaml
```diff
@@ -91,9 +91,70 @@ RedHat:
 {% endif %}
 
 Suse:
+  pkg_repo:
+    name: pgdg-sles-{{ release }}
+    humanname: PostgreSQL {{ repo.version }} $releasever - $basearch
+    #Suse repos are suffixed 'suse/sles-12-$basearch' by upstream community convention.
+    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/suse/sles-12-$basearch'
+    gpgkey:  'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/suse/sles-12-$basearch/repodata/repomd.xml'
```
myoung34: this isnt a gpg key, this is a yum repo metadata xml file

postgres/osmap.yaml
```diff
@@ -91,9 +91,70 @@ RedHat:
...
+    #future: fix suse "4-Signatures public key is not available" issue and enable gpgcheck.
+    gpgcheck: 0
```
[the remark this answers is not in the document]
noelmcloughlin: I agree its not nice - I'm looking for agreement to raise as "ISSUE" and fix afterwards.

postgres/osmap.yaml
```diff
@@ -91,9 +91,70 @@ RedHat:
 {% endif %}
 
 Suse:
+  pkg_repo:
+    name: pgdg-sles-{{ release }}
+    humanname: PostgreSQL {{ repo.version }} $releasever - $basearch
+    #Suse repos are suffixed 'suse/sles-12-$basearch' by upstream community convention.
+    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/suse/sles-12-$basearch'
```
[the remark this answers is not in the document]
noelmcloughlin: I cannot use '$releasever' because packages target Suse family (SUSE 12+, OpenSUSE 42+), so the convention today seems to be the hardcoding of "sles-12".  Using '$releasever' breaks the formula for Suse family except Sles 12.

postgres/osmap.yaml
noelmcloughlin: Good spot, thanks.

postgres/osmap.yaml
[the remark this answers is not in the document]
noelmcloughlin: Yes I done some troubleshooting. But @myoung34 has identified the problem.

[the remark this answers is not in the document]
vutny: You just need to provide correct public key URL, and I think everything would just "magically" work :)

postgres/osmap.yaml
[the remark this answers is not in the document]
vutny: Yeah, that's true, I understand. My point was that the installation could be broken on other SUSE variants which is not currently supported by upstream: SLES 11, OpenSuse 13 or Tumbleweed.
In such cases it's better to fail fast. But that also introduces complexities.
I think this is fine for now.

postgres/osmap.yaml
vutny: The SUSE section does need to be full copy of RedHat just for the sake of artificial consistency.
These options `bin_dir`, `client_bins` and `server_bins` serve a concrete purpose: allow subsequent states and `postges` modules to properly discover binaries and be able to do the job.

Other distros don't have them, because they don't need to. That's why it is configurable.
The necessary thing should be already done by installed packages, so the formula must not duplicate this.

pillar.example
vutny: Better to test your changes on other platforms as well, because you change shared things, and `pillar.example` has been used in existent test. That's not perfect, and the change is actually meaningful, but we need to understand an impact.

postgres/osmap.yaml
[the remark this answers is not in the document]
noelmcloughlin: Duh, need to wear my reading glasses.

postgres/osmap.yaml
vutny: Actually, it was a rhetoric question :)
https://software.opensuse.org/package/postgresql

postgres/osmap.yaml
noelmcloughlin: This alternatives cod is legacy. Perhaps in the past the packages did not already do this. 
 
I will investigate and raise common issue on RedHat/Suse.

vutny: And Suse definitely doesn't require that.

pillar.example
noelmcloughlin: I've already been testing formula on other distributions - see #171 #170 #164.

Only regression is #172 which is something you mentioned regarding pkgs_extra()

vutny: I assume this affects Ubuntu and the existing test based on it.

myoung34:
I see that there's a `.kitchen.yml` file. can we get @aboe76 to enable it to prevent some regressions?

pillar.example
myoung34: Also this is just a `pillar.example`, just leave the values as is to help people understand layout. We only get one pillar.example and support (unofficially) any number of distros so it doesnt need to change with every iteration of what distro was last used against it.

postgres/osmap.yaml
myoung34: There has to be a better way to handle this than to cargo cult large blocks that are similar between os families

pillar.example
vutny: That's exactly what I've been trying to express 👍 But having some working example that fits the most popular ones would be great.

noelmcloughlin: Agreed. Will action.

postgres/osmap.yaml
noelmcloughlin: I'm sorry, okay thanks.

postgres/osmap.yaml
noelmcloughlin: Thanks for pointing me in right way here. Surely RedHat has same problem.  

Agreed - I tested and all except two are already handled by upstream packages.

postgres/osmap.yaml
vutny: I have found that starting from PostgreSQL 9.6 they ship valid repo key within the repo:
https://download.postgresql.org/pub/repos/zypp/9.6/suse/sles-12-x86_64/repodata/repomd.xml.key

I think we could leave 9.5 and versions below as is, since user is able to override GPG check in Pillar.

vutny:
Hi @noelmcloughlin 

Could you please update the status of this PR? Would you have some time or any plans to work on it further according to review comments?
Thanks for your contributions anyway!

noelmcloughlin:
Hey @vutny 
To recap, the original PR proposes adding support for OpenSUSE without Hash-checking.  PR reviewers preferred full feature rather than incremental improvement.  However, full feature is blocked by  issues mentioned in #173.  **The correct action here is to close the PR** and I should refactor a new solution acceptable to community.

Thank you for reviewing!!!!!  I will followup on #173 and your comments in this review in a new PR as soon as I make some time.

vutny:
@noelmcloughlin Awesome, thanks. So let's keep this PR open as a reminded until you'd have another one refactored. Then we close it out in favor of that.

noelmcloughlin:
Hi @vutny I'll pick this up again as next priority, been hectic with other stuff. thx

noelmcloughlin:
Rebased as https://github.com/saltstack-formulas/postgres-formula/pull/176

Issue #172, opened 2017-09-25
State postgres_extension-uuid-ossp fails


noelmcloughlin:
Needs investigation. Pillars based on defaults in pillar.example. OpenSUSE Leap 42

```
[INFO  ]  Failed to create ext: uuid-ossp
[ERROR] Failed to install extension uuid-ossp
```

noelmcloughlin:
This was a formula rerun.

noelmcloughlin:
The root cause may be pillar related. @vutny

noelmcloughlin:
Caused by missing `postgresql-contrib` package. See #176

noelmcloughlin:
Here is one solution - only install postgres-contrib - when extensions are defined.  This package contains "extensions and addons not part of core" according to packaging info.
```
{%- if postgres.extensions %}
  {%- set pkgs = [postgres.pkg] + [postgres.pkg_contrib] + postgres.pkgs_extra %}
{%- else %}
  {%- set pkgs = [postgres.pkg] + postgres.pkgs_extra %}
{% endif %}
```

Other solution suggestions welcome.

noelmcloughlin:
No time to look at this issue. closing for now.

Issue #160, opened 2017-09-14
Upstream Packaging:  /bin/bash: initdb: command not found


noelmcloughlin:
Deploy of postfix 9.6 on Ubuntu fails.

Replication
- Fresh vanilla installation of Ubuntu 16/17 & SALT 2017.17
- Change version "9.3" to "9.6" in pillars populated from pillar.example
https://github.com/saltstack-formulas/postgres-formula/blob/master/pillar.example

Output:
>
----------
          ID: postgresql-client-libs
    Function: pkg.installed
      Result: True
     Comment: 2 targeted packages were installed/updated.
     Started: 22:29:56.822971
    Duration: 14458.556 ms
     Changes:   
              ----------
              libpq-dev:
                  ----------
                  new:
                      9.6.4-0ubuntu0.17.04.1
                  old:
              libpq5:
                  ----------
                  new:
                      9.6.4-0ubuntu0.17.04.1
                  old:
              postgresql-client:
                  ----------
                  new:
                      1
                  old:
              postgresql-client-9.6:
                  ----------
                  new:
                      9.6.4-0ubuntu0.17.04.1
                  old:
              postgresql-client-common:
                  ----------
                  new:
                      179
                  old:
----------
          ID: postgresql-server
    Function: pkg.installed
      Result: True
     Comment: 3 targeted packages were installed/updated.
     Started: 22:30:11.289041
    Duration: 31987.267 ms
     Changes:   
              ----------
              postgresql-9.6:
                  ----------
                  new:
                      9.6.4-0ubuntu0.17.04.1
                  old:
              postgresql-common:
                  ----------
                  new:
                      179
                  old:
              postgresql-contrib:
                  ----------
                  new:
                      9.6+179
                  old:
              postgresql-contrib-9.6:
                  ----------
                  new:
                      9.6.4-0ubuntu0.17.04.1
                  old:
              postgresql-plpython:
                  ----------
                  new:
                      1
                  old:
              postgresql-plpython-9.6:
                  ----------
                  new:
                      9.6.4-0ubuntu0.17.04.1
                  old:
              sysstat:
                  ----------
                  new:
                      11.4.3-1
                  old:
----------
          ID: postgresql-cluster-prepared
    Function: cmd.run
        Name: initdb --pgdata=/var/lib/pgsql/data
      Result: False
     Comment: Command "initdb --pgdata=/var/lib/pgsql/data" run
     Started: 22:30:43.280835
    Duration: 289.877 ms
[...]

PR #176, opened 2018-01-14
Support for upstream postgresql.org zypp repo

noelmcloughlin:
This PR is an enhancement to allow the [upstream repo](https://download.postgresql.org/pub/repos/zypp/) to be used for SLES and OpenSUSE. 

This rebased commit replaces PR https://github.com/saltstack-formulas/postgres-formula/pull/169

Tested successfully on version '9.6' and '10'. Failed with 9.5 due to upstream repo problem.

```
postgres:
  use_upstream_repo: True
  version: '10'
  pkgs_extra:
  - postgresql10-contrib
  - postgresql10-plpython

base:
  '*':
    - postgres
    - postgres.client
    - postgres.manage
    - postgres.python
    - postgres.server
    - postgres.server.image
    - postgres.upstream

---------
          ID: postgresql-repo
    Function: pkgrepo.managed
        Name: pgdg-sles-10
      Result: True
     Comment: Configured package repo 'pgdg-sles-10'
     Started: 14:32:38.642363
    Duration: 47.584 ms
     Changes:   
----------
          ID: postgresql-repo
    Function: cmd.run
        Name: zypper --gpg-auto-import-keys refresh --force
      Result: True
     Comment: Command "zypper --gpg-auto-import-keys refresh --force" run
     Started: 14:32:38.691449
    Duration: 8691.113 ms
     Changes:   
              ----------
              pid:
                  13510
              retcode:
                  0
              stderr:
              stdout:
                  Forcing raw metadata refresh
                  Retrieving repository 'pgdg-sles-10' metadata [...done]
                  Forcing building of repository cache
                  Building repository 'pgdg-sles-10' cache [....done]
                  Forcing raw metadata refresh
                  Retrieving repository 'openSUSE-Leap-42.3-Non-Oss' metadata [.done]
                  Forcing building of repository cache
                  Building repository 'openSUSE-Leap-42.3-Non-Oss' cache [....done]
                  Forcing raw metadata refresh
                  Retrieving repository 'openSUSE-Leap-42.3-Oss' metadata [..done]
                  Forcing building of repository cache
                  Building repository 'openSUSE-Leap-42.3-Oss' cache [....done]
                  Forcing raw metadata refresh
                  Retrieving repository 'openSUSE-Leap-42.3-Update' metadata [..done]
                  Forcing building of repository cache
                  Building repository 'openSUSE-Leap-42.3-Update' cache [....done]
                  Forcing raw metadata refresh
                  Retrieving repository 'openSUSE-Leap-42.3-Update-Non-Oss' metadata [..done]
                  Forcing building of repository cache
                  Building repository 'openSUSE-Leap-42.3-Update-Non-Oss' cache [....done]
                  Forcing raw metadata refresh
                  Retrieving repository 'Salt releases for SLE-based SUSE products (openSUSE_Leap_42.3)' metadata [..done]
                  Forcing building of repository cache
                  Building repository 'Salt releases for SLE-based SUSE products (openSUSE_Leap_42.3)' cache [....done]
                  All repositories have been refreshed.
----------
          ID: postgresql-server
    Function: pkg.installed
      Result: True
     Comment: All specified packages are already installed
     Started: 14:32:47.389718
    Duration: 314.426 ms
     Changes:   
----------
    Function: cmd.run
        Name: initdb --pgdata='/var/lib/pgsql/10/data'
      Result: True
     Comment: unless execution succeeded
     Started: 14:32:47.711825
    Duration: 45.167 ms
     Changes:   
----------
          ID: postgresql-config-dir
    Function: file.directory
        Name: /var/lib/pgsql/10/data
      Result: True
     Comment: Directory /var/lib/pgsql/10/data is in the correct state
              Directory /var/lib/pgsql/10/data updated
     Started: 14:32:47.758883
    Duration: 0.474 ms
     Changes:   
----------
          ID: postgresql-pg_hba
    Function: file.managed
        Name: /var/lib/pgsql/10/data/pg_hba.conf
      Result: True
     Comment: File /var/lib/pgsql/10/data/pg_hba.conf is in the correct state
     Started: 14:32:47.759573
    Duration: 162.051 ms
     Changes:   
----------
          ID: postgresql-running
    Function: service.running
        Name: postgresql-10
      Result: True
     Comment: The service postgresql-10 is already running
     Started: 14:32:47.914996
    Duration: 22.87 ms
     Changes:   
----------
          ID: postgresql-client-libs
    Function: pkg.installed
      Result: True
     Comment: All specified packages are already installed
     Started: 14:32:47.945155
    Duration: 19.896 ms
     Changes:   
----------
          ID: postgres-reload-modules
    Function: test.nop
      Result: True
     Comment: Success!
     Started: 14:32:47.965238
    Duration: 0.398 ms
     Changes:   
----------
          ID: postgresql-python
    Function: pkg.installed
        Name: python-psycopg2
      Result: True
     Comment: All specified packages are already installed
     Started: 14:32:47.965724
    Duration: 22.854 ms
     Changes:   
----------
          ID: postgresql-start
    Function: test.show_notification
      Result: True
     Comment: The 'postgres:bake_image' Pillar is disabled (set to 'False').
     Started: 14:32:47.988690
    Duration: 0.228 ms
     Changes:   

Summary for local
-------------
Succeeded: 11 (changed=1)
Failed:     0
-------------
Total states run:     11
Total run time:    9.334 s
```

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -1,19 +1,20 @@
 postgres:
-  # Set True to configure upstream postgresql.org repository for YUM or APT
+  # UPSTREAM REPO
+  # Set True to configure upstream postgresql.org repository for YUM/APT/ZYPP
   use_upstream_repo: False
-  # Version to install from upstream repository
-  version: '9.3'
+  # Version to install from upstream repository (if upstream_repo: True)
+  version: '9.6'
 
-  # These are Debian/Ubuntu specific package names
-  pkg: 'postgresql-9.3'
-  pkg_client: 'postgresql-client-9.3'
-
-  # Additional packages to install with PostgreSQL server,
-  # this should be in a list format
+  # PACKAGE
+  # These pillars are typically never required.
+  # pkg: 'postgresql'
+  # pkg_client: 'postgresql-client'
+  # service: postgresql
   pkgs_extra:
     - postgresql-contrib
     - postgresql-plpython
 
+  # POSTGRES
   # Append the lines under this item to your postgresql.conf file.
   # Pay attention to indent exactly with 4 spaces for all lines.
   postgresconf: |
@@ -47,9 +48,6 @@ postgres:
   {%- if salt['status.time']|default(none) is callable %}
   config_backup: ".backup@{{ salt['status.time']('%y-%m-%d_%H:%M:%S') }}"
   {%- endif %}
-
-  # PostgreSQL service name
-  service: postgresql
 
   {%- if grains['init'] == 'unknown' %}
 
--- a/postgres/osmap.yaml
+++ b/postgres/osmap.yaml
@@ -91,9 +91,43 @@ RedHat:
 {% endif %}
 
 Suse:
+  pkg_repo:
+    name: pgdg-sles-{{ release }}
+    humanname: PostgreSQL {{ repo.version }} $releasever - $basearch
+    #Using sles-12 upstream repo for opensuse
+    baseurl: 'https://download.postgresql.org/pub/repos/zypp/{{ repo.version }}/suse/sles-12-$basearch'
+    key_url: 'https://download.postgresql.org/pub/repos/zypp/{{ repo.version }}/suse/sles-12-$basearch/repodata/repomd.xml.key'
+    gpgcheck: 1
+    gpgautoimport: True
+
+{% if repo.use_upstream_repo %}
+  {# Pillars needed are 'use_upstream_repo: True' and 'version: n.n'. #}
+  {# Avoid setting package names as pillars, as may corrupt postgres. #}
+  {% set lib_dir = '/var/lib/pgsql/' ~ repo.version ~ '/data' %}
+
+  pkg: postgresql{{ release }}-server
+  pkg_client: postgresql{{ release }}
+  conf_dir: {{ lib_dir }}
+  service: postgresql-{{ repo.version }}
+
+  #This is postgresql-libs in defaults.yml but upstream is libpqxx
+  pkg_libpq_dev: libpqxx
+  pkg_dev: postgresql{{ release }}-devel
+  pkg_libs: postgresql{{ release }}-libs
+
+  prepare_cluster:
+    #Full path needed as initdb is NOT 'cross version compatible' binary
+    command: /usr/pgsql-{{ repo.version }}/bin/initdb --pgdata='{{ lib_dir }}'
+    test: test -f '{{ lib_dir }}/PG_VERSION'
+
+{% else %}
+
   pkg: postgresql-server
   pkg_client: postgresql
-  pkg_libpq_dev: postgresql
+  pkg_libpq_dev: libqpxx
+  pkg_libs: postgresql-libs
+
+{% endif %}
 
 MacOS:
   service: postgresql
@@ -108,6 +142,5 @@ MacOS:
     test: test -f /usr/local/var/postgres/PG_VERSION
     user: _postgres
     group: _postgres
-
 
 # vim: ft=sls


noelmcloughlin:
Hi @vutny Can you merge this if you are happy please?

noelmcloughlin:
Is anyone able to look at this?

pillar.example
```diff
@@ -1,18 +1,18 @@
...
   use_upstream_repo: False
   # Version to install from upstream repository
-  version: '9.3'
+  version: '9.6'
 
   # These are Debian/Ubuntu specific package names
-  pkg: 'postgresql-9.3'
-  pkg_client: 'postgresql-client-9.3'
+  pkg: 'postgresql-9.6'
+  pkg_client: 'postgresql-client-9.6'
```
vutny: As the comment above said, these values are specific to Ubuntu, 14.04 to be precise.
This is because the `pillar.example` is used in `test-kitchen` within the formula.
That's kind of legacy, but still serves as smoke test for regressions.

Since now we have good support for package names mapping for various distros, I think it could be safely commented out.

noelmcloughlin: Can we go further and remove all package names from pillar.example?

Executing the formula with pillars `use_upstream_repo: True` with `pkg_extras: ['postgresql-contrib', 'postgresql-plpython']` corrupts the postgres installation for example. They are not needed imho.

noelmcloughlin: And remove`service` pillar too.    And pkgs_extra example pillar implies `postgres-contrib` is optional "extra package", but the `postgres.prepare_cluster.command` fails (cannot find initdb) if package is not installed, implying its needed.

pillar.example
```diff
@@ -1,18 +1,18 @@
...
 
   # Additional packages to install with PostgreSQL server,
   # this should be in a list format
-  pkgs_extra:
-    - postgresql-contrib
-    - postgresql-plpython
+  #pkgs_extra:
+  #  - postgresql-contrib
```
vutny: Regarding your comment above. If we're talking about Ubuntu, `postgresql-contrib` is definitely optional package, it does not contain `initdb` binary.
https://packages.ubuntu.com/trusty/amd64/postgresql-contrib-9.3/filelist
https://packages.ubuntu.com/xenial/amd64/postgresql-contrib-9.5/filelist

I think the issue with missing `initdb` lies somewhere else, maybe in our beloved alternatives?
So commenting out these extras is a good thing.

pillar.example
vutny: You're right, the package and service definitions in Pillar are mostly unneeded.
But they could be useful when trying to install something on a latest-greatest and still unsupported distro.
I think it would be better to add a comment that would mention they are unnecessary in most cases.

pillar.example
noelmcloughlin: Yeah, `postgres-contrib` creates alternatives. The formula requires initdb in PATH (see below) or state failures occur.
```
./postgres/defaults.yaml:    command: initdb --pgdata=/var/lib/pgsql/data
./postgres/osmap.yaml:    command: initdb -D /var/lib/postgres/data
./postgres/osmap.yaml:    command: initdb --pgdata='{{ data_dir }}'
./postgres/osmap.yaml:    - initdb
./postgres/osmap.yaml:    command: initdb --pgdata='{{ lib_dir }}'
./postgres/osmap.yaml:    command: initdb -D /Users/{{ repo.user }}/Library/AppSupport/postgres_{{ repo.use_upstream_repo }}
```

Do you know how `initdb` gets added to PATH without `postgres-contrib` package in this formula?  The redhat os mapping invokes saltstack to create alternatives, so the package must have been missing when that feature was missing.

Back to the specific comment - I would prefer to remove pkgs_extra pillar example, because it corrupted my installation once or twice, this is bad onboarding experience for new users. Otherwise commenting is definitely a good thing - I agree and will implement.

pillar.example
noelmcloughlin: I'll think about this, thanks.

pillar.example
vutny: This is because of the `postgres-contrib` is just a metapackage that *also* install a full PostgreSQL server + client of particular version available in the distro by dependencies. And there we have all the alternatives installation code in the downstream.
So if you have a wrong or empty value in the `pkg` Pillar value, the `postgres-contrib` just does the right thing.

This is just a very handy coincidence, that does not make `postgres-contrib` package mandatory.

pillar.example
vutny: Sure, just for now I think we should comment them out to make `test-kitchen` work, but add a proper description for these parameter later.
Or use generic ones like `postgresql` and `postgresql-client`. It will make transition to newer distro version much smoother as well.

pillar.example
noelmcloughlin: Okay. In this PR I explicitly fix the INITDB problem (#160) via salt.alternatives.install. As documented in #160, upstream packages do not add initdb to PATH by design. 

Here is my verified solution - I am unaware of any postgres-defined solution.
```
...
 +  prepare_cluster:
 +    #Full path needed as initdb is NOT 'cross version compatible' binary
 +    command: /usr/pgsql-{{ release }}/bin/initdb --pgdata='{{ lib_dir }}'
 +    test: test -f '{{ lib_dir }}/PG_VERSION'
 +
 +  # Handles etc alternatives for 'cross version compatible' binaries
 +  pkgs_extra: ['postgresql{{ release }}-contrib',]
 +
 +{% else %}
```
The solution fails when pillars overrides the verified os mapping values. 

If this solution is acceptable to you, let me know.

vutny: Yeah. That is a separate issue appearing exclusively when installing from upstream.
We will handle this. Thank you.

postgres/client.sls
```diff
@@ -23,11 +23,14 @@ postgresql-client-libs:
     - refresh: True
     - require:
       - pkgrepo: postgresql-repo
+       {% if grains.os_family == 'Suse' %}
+      - cmd: postgresql-repo
+       {% endif %}
```
vutny: I think here would be better to depend on the whole `postgres.upstream` SLS to reduce verbose Jinja statements:
```
    ...
    - require:
      - sls: postgres.upstream

File: postgres/client.sls [hunk 4]
```diff
@@ -23,11 +23,14 @@ postgresql-client-libs:
...
 {%- endif %}
 
 {%- if 'bin_dir' in postgres %}
 
-# Make client binaries available in $PATH
+  # Make client binaries available in $PATH
```
vutny: No need for indenting these comments, since they're part of rendered YAML, not Jinja.
It would look better in debug log when they are aligned with states they describe.

postgres/client.sls
```diff
@@ -23,6 +23,9 @@ postgresql-client-libs:
     - refresh: True
     - require:
       - pkgrepo: postgresql-repo
+       {% if grains.os_family == 'Suse' %}
```
vutny: I meant, no need for `{% if grains... %}` statements and `pkgrepo` requisite above as well 😃

noelmcloughlin: no problem. check now.

postgres/upstream.sls
```diff
@@ -4,11 +4,17 @@
...
   {%- if postgres.use_upstream_repo -%}
-
 # Add upstream repository for your distro
+
 postgresql-repo:
   pkgrepo.managed:
     {{- format_kwargs(postgres.pkg_repo) }}
+    {% if grains.os_family == 'Suse' %}
+  cmd.run:
+    - name: zypper --gpg-auto-import-keys refresh --force
```
vutny: This is in connection with what we've discussed about requisites for the `pkg` state.
I know it is not obvious from the Salt docs, but `pkgrepo.managed` state supports special option `gpgautoimport` when target package manager is Zypper.
No need for messing with grains here. I'll show how simple it could be done above at `osmap.yaml` file.

postgres/osmap.yaml
```diff
@@ -91,9 +91,45 @@ RedHat:
 {% endif %}
 
 Suse:
+  pkg_repo:
+    name: pgdg-sles-{{ release }}
+    humanname: PostgreSQL {{ repo.version }} $releasever - $basearch
+    #Using sles-12 upstream repo for opensuse
+    baseurl: 'https://download.postgresql.org/pub/repos/zypp/{{ repo.version }}/suse/sles-12-$basearch'
+    key_url: 'https://download.postgresql.org/pub/repos/zypp/{{ repo.version }}/suse/sles-12-$basearch/repodata/repomd.xml.key'
+    gpgcheck: 1
```
vutny: Just add
```yaml
    gpgautoimport: True
```
 
bellow, and importing GPG key from the upstream will start to magically work. No need for additional `cmd.run` state in the `upstream.sls`.

postgres/upstream.sls
noelmcloughlin: How did you know that ;-)   I did check the [docs](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkgrepo.html).

postgres/osmap.yaml
noelmcloughlin: Thanks - done - let me know when your PR comments are done and I'll do some retesting on Suse to make sure there is no regression.

postgres/osmap.yaml
```diff
@@ -91,9 +91,46 @@ RedHat:
...
+  pkg_dev: postgresql{{ release }}-devel
+  pkg_libs: postgresql{{ release }}-libs
+
+  prepare_cluster:
+    #Full path needed as initdb is NOT 'cross version compatible' binary
+    command: /usr/pgsql-{{ release }}/bin/initdb --pgdata='{{ lib_dir }}'
+    test: test -f '{{ lib_dir }}/PG_VERSION'
+
+  # Handles etc alternatives for 'cross version compatible' binaries
+  pkgs_extra: ['postgresql{{ release }}-contrib',]
```
vutny: I've checked out the `postgresql10-contrib-10.1-1PGDG.sles12.x86_64.rpm` and appears it has nothing to do with alternatives at all.
I think we should disable extra packages installation by default, but may add a comment to `pillar.example` file mentioning how typically they called if having `use_upstream_repo` enabled.

noelmcloughlin: Thank you for checking. I fully agree extra package installation can be disabled by default for general OS mapping/defaults/pillar.example.

I will retest with Zypp upstream and if not needed will disabled the variable for Suse.

postgres/osmap.yaml
```diff
@@ -91,9 +91,46 @@ RedHat:
...
+{% if repo.use_upstream_repo %}
+  {# Pillars needed are 'use_upstream_repo: True' and 'version: n.n'. #}
+  {# Avoid setting package names as pillars, as may corrupt postgres. #}
+  {% set lib_dir = '/var/lib/pgsql/' ~ repo.version ~ '/data' %}
+
+  pkg: postgresql{{ release }}-server
+  pkg_client: postgresql{{ release }}
+  conf_dir: {{ lib_dir }}
+   {% if release|int == 10 %}
+  service: postgresql-{{ release }}
```
vutny: I've checked out `postgresql96-server-9.6.6-1PGDG.sles12.x86_64.rpm` package, and there the service name follows the same pattern. I think we should omit release version check above.

noelmcloughlin:
Hi @vutny, I have done testing and few updates were needed to consistently pass verification. Please review again when you have time.   Thanks.

noelmcloughlin:
Hi @vutny can I merge this?

postgres/server/init.sls
```diff
@@ -15,7 +15,11 @@ include:
 
 {%- endif %}
 
-{%- set pkgs = [postgres.pkg] + postgres.pkgs_extra %}
+{%- if postgres.extensions %}
+  {%- set pkgs = [postgres.pkg] + [postgres.pkg_contrib] + postgres.pkgs_extra %}
```
vutny: This statement looks very verbose and unnecessary...
It declares strait dependency between `extensions` setting in Pillar and newly introduced `pkg_contrib` value, which in turn seems to be applicable only on SUSE distros.

* What if desired extensions are not shipped with package defined in `pkg_contrib`?
* I see it was set empty (evaluates to `None`) for other platforms. Will it work if some extensions have been set?

It looks little bit confusing, and is not described anywhere.
I suggest to keep the things as is: if user needs some extension, he or she should provide extra packages via `pkgs_extra` or install them another way.

noelmcloughlin: Thanks @vutny The extensions have dependency on *some* package in PG and its gray area causing issues like #172 - suggestions welcomed there.  I agree to remove this from PR code.

noelmcloughlin:
I've removed the if postgres.extensions check.

postgres/osmap.yaml
```diff
@@ -91,9 +91,45 @@ RedHat:
...
+  pkg: postgresql{{ release }}-server
+  pkg_client: postgresql{{ release }}
+  conf_dir: {{ lib_dir }}
+  service: postgresql-{{ repo.version }}
+
+  #This is postgresql-libs in defaults.yml but upstream is libpqxx
+  pkg_libpq_dev: libpqxx
+  pkg_dev: postgresql{{ release }}-devel
+  pkg_libs: postgresql{{ release }}-libs
+  pkg_contrib: postgresql{{ release }}-contrib
```
vutny: Here and below, the `pkg_contrib` parameter becomes unused. So no need to keep it.

postgres/osmap.yaml
```diff
@@ -91,9 +91,45 @@ RedHat:
...
+    test: test -f '{{ lib_dir }}/PG_VERSION'
+
+{% else %}
+
   pkg: postgresql-server
   pkg_client: postgresql
-  pkg_libpq_dev: postgresql
+  pkg_libpq_dev: libpq5
+  pkg_libs: postgresql-libs
+  pkg_contrib: postgresql-contrib
```
vutny: You may put the `postgresql-contrib` back into the proper section in `pillar.example` file with comment saying this package could be a requirement for installing some extensions.

postgres/osmap.yaml
```diff
@@ -91,9 +91,45 @@ RedHat:
...
+    #Full path needed as initdb is NOT 'cross version compatible' binary
+    command: /usr/pgsql-{{ repo.version }}/bin/initdb --pgdata='{{ lib_dir }}'
+    test: test -f '{{ lib_dir }}/PG_VERSION'
+
+{% else %}
+
   pkg: postgresql-server
   pkg_client: postgresql
-  pkg_libpq_dev: postgresql
+  pkg_libpq_dev: libpq5
```
vutny: I found that OpenSuse also has C++ bindings package named [libpqxx](https://software.opensuse.org/package/libpqxx) similar to the upstream.

postgres/osmap.yaml
noelmcloughlin: I don't want to. Pillar handling exposes production users to unacceptable risk.

- Handling packages via osmap.yaml pattern is good.
- Handling packages **only** via pillars (i.e. postgresql_contrib) is bad pattern.
-- Missing Postgresql-contrib package causes bug #172.
-- Specifying wrong contrib_pkg pillar corrupts package manager, the PG installation, risks data loss. The PG schema may get upgraded too. This is all bad.
- And trying to handle #172 via comments is wrong approach.

postgres/osmap.yaml
noelmcloughlin: Thanks for checking. Should libpqxx value replace 'libpq5'?

vutny: Yeah, exactly as in the upstream case to have everything consistent.

postgres/osmap.yaml
vutny: I mostly agree with all your points.
My main point is that `pkg_contrib` have not been used within current state of your changes, so having this in defaults and OS mapping files is meaningless right now. It just does nothing.
And this does not really help to solve issue #172 taking into an account broad scope of distributions supported by the formula.
I will try to put my thoughts there. Let's concentrate on bringing in initial basic support for SUSE systems (maybe less functional), but without breaking things which already work.

postgres/osmap.yaml
noelmcloughlin: Where did libpq5 come from?  Cut and paste? maybe by me? ;-(

postgres/osmap.yaml
noelmcloughlin: I'll remove these entries from osmap.yaml but don't want pkg_contrib in pillar.example. Apart from the reasons above, I've done lots of verification on this formula past few weeks across many scenarios and I just don't agree with packages in pillars anymore - there are too many use cases.  But, I do want the PR merged, so tell me if you want pillar.example restored.

postgres/defaults.yaml
```diff
@@ -5,6 +5,7 @@ postgres:
   version: '9.5'
   pkg: postgresql
   pkgs_extra: []
+  pkg_contrib:
```
vutny: This one has been forgotten.

noelmcloughlin: Got it!! Sorry.

postgres/osmap.yaml
```diff
@@ -91,9 +91,43 @@ RedHat:
...
+  service: postgresql-{{ repo.version }}
+
+  #This is postgresql-libs in defaults.yml but upstream is libpqxx
+  pkg_libpq_dev: libpqxx
+  pkg_dev: postgresql{{ release }}-devel
+  pkg_libs: postgresql{{ release }}-libs
+
+  prepare_cluster:
+    #Full path needed as initdb is NOT 'cross version compatible' binary
+    command: /usr/pgsql-{{ repo.version }}/bin/initdb --pgdata='{{ lib_dir }}'
```
vutny: This is generally OK. But I would recommend you to consider providing `bin_dir` and `*_bins` parameters similar to RedHat section for `postgres.manage` state to work on Suse out of the box as well.
But that could be added later.

pillar.example
```diff
@@ -1,19 +1,19 @@
...
-
+  # Version to install from upstream repository (if upstream_repo: True)
+  version: '9.6'
+
+  # PACKAGE
+  # These pillars are typically never required.
+  # pkg: 'postgresql'
+  # pkg_client: 'postgresql-client'
+  # service: postgresql
+  #pkgs_extra:
```
vutny: This `pkg_extra` parameter still concerns me.

I thought we reached agreement that `pillar.example` file could not be used as valid source for any possible platform which the formula supports. Hard-coding some package names is bad thing too.
But now this breaks the test kitchen run.
We have two options:
* Uncomment `pkgs_extra` and add `postgresql-contrib` there mentioned this could be used only with `use_upstream_repo: False`.
* ...or replace `uuid-ossp` extension with `plpgsql`, that comes with PG server. The schema name also should be changed to some `custom_schema` for consistency.

noelmcloughlin: I have reinstated `pkgs_extra` as per original pillar.example.   

Yeah, I'm unconvinced about pillar.example. The README and pillar.example are both delivered artifacts, responsible for onboarding new formula users (perhaps in a 1 week agile sprint), and guiding adoption for happy and unhappy paths. test kitchen is a platform too.   **On the other hand, I agree the pillar.example must have limited scope**, which aligns with your stated view.

postgres/osmap.yaml
noelmcloughlin: https://github.com/saltstack-formulas/postgres-formula/issues/189

noelmcloughlin: Also https://github.com/saltstack-formulas/postgres-formula/issues/190 captures problem regarding package names in pillar data.

noelmcloughlin:
Excellent work reviewing this, thanks.

aboe76:
@noelmcloughlin and @vutny thanks for all the hard work.

PR #177, opened 2018-01-14
W/A hardcoding of debian alternatives priority

noelmcloughlin:
The current hardcoding of '`postgres.alternatives.install.priority=30`', breaks formula reruns.

`-    - priority: 30`

This PR is a workaround solution, by exposing the value as a configurable pillar.

`+    - priority: {{ postgres.linux.altpriority|int }}`

**Implementation decisions**

1. To mimic existing behaviour the alternatives.priority value of '30' is retained by defaults.yaml.
2. Users can disable Debian Alternatives - new feature - by setting '0' as pillar value (that would be better formula default IMHO but lets go one step at a time).

### DIFF
--- a/postgres/client.sls
+++ b/postgres/client.sls
@@ -8,14 +8,11 @@
 {%- endfor %}
 
 {%- if postgres.use_upstream_repo %}
-
 include:
   - postgres.upstream
-
 {%- endif %}
 
 # Install PostgreSQL client and libraries
-
 postgresql-client-libs:
   pkg.installed:
     - pkgs: {{ pkgs }}
@@ -25,23 +22,19 @@ postgresql-client-libs:
       - pkgrepo: postgresql-repo
 {%- endif %}
 
-{%- if 'bin_dir' in postgres %}
-
-# Make client binaries available in $PATH
-
-  {%- for bin in postgres.client_bins %}
-
-    {%- set path = salt['file.join'](postgres.bin_dir, bin) %}
+# Alternatives system. Make client binaries available in $PATH
+{%- if 'bin_dir' in postgres and postgres.linux.altpriority %}
+    {%- for bin in postgres.client_bins %}
+      {%- set path = salt['file.join'](postgres.bin_dir, bin) %}
 
 {{ bin }}:
   alternatives.install:
     - link: {{ salt['file.join']('/usr/bin', bin) }}
     - path: {{ path }}
-    - priority: 30
+    - priority: {{ postgres.linux.altpriority }}
     - onlyif: test -f {{ path }}
     - require:
       - pkg: postgresql-client-libs
 
-  {%- endfor %}
-
+    {%- endfor %}
 {%- endif %}
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -41,3 +41,7 @@ postgres:
   databases: {}
   schemas: {}
   extensions: {}
+
+  linux:
+    #Alternatives system are disabled by a 'altpriority=0' pillar.
+    altpriority:
--- a/pillar.example
+++ b/pillar.example
@@ -14,6 +14,10 @@ postgres:
     - postgresql-contrib
     - postgresql-plpython
 
+  #'Alternatives system' priority incremental. 0 disables feature.
+  linux:
+    altpriority: 30
+
   # Append the lines under this item to your postgresql.conf file.
   # Pay attention to indent exactly with 4 spaces for all lines.
   postgresconf: |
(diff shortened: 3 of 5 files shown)


noelmcloughlin:
Would anyone be able to review?

EvaSDK:
I don't think debian alternatives makes sense on os_family other than Debian so I'm not sure the conditional is right.

noelmcloughlin:
I think `debian alternatives` is a historical name. Its called '`alternatives`' on RedHat for example-

> alternatives is a tool for managing different software packages that provide the same functionality. Red Hat Enterprise Linux uses alternatives to ensure that only one Java Development Kit (JDK) is set as the system default at one time.

noelmcloughlin:
BTW, this PR does not add alternatives feature, just removes a hardcoding of 30 priority.   

There is a pre-existing check for os_family which do not support 'debian alternatives'.

` {% if grains.os_family not in ('Arch',) %}`

EvaSDK:
Gentoo has eselect which is somewhat similar but has its own module https://docs.saltstack.com/en/latest/ref/states/all/salt.states.eselect.html. Not sure alternatives would work there.

noelmcloughlin:
Good point @EvaSDK  Could you raise issue?  BTW, such problems exist as legacy technical debt.

noelmcloughlin:
Hey @EvaSDK I updated commit for your comments.  Was thinking this PR assists future eselect because the pillar is reusable for that feature. Probaby eselect has priorities too?

postgres/client.sls
```diff
@@ -25,23 +22,27 @@ postgresql-client-libs:
       - pkgrepo: postgresql-repo
 {%- endif %}
 
+# Alternatives system
 {%- if 'bin_dir' in postgres %}
+  {% if postgres.linux.altpriority|int > 0 %}
+    {% if grains.os_family not in ('Arch', 'Gentoo', 'MacOS', 'Windows',) %}
```
vutny: You don't need to hard-code the list of distros in the state.
The top-level `if` statement checks if `bin_dir` parameter has been defined. And it is explicitly defined only for RedHat in `osmap.yaml` for now. It does not exist for any other system.

Same applies to the `server` state.

noelmcloughlin: Thats correct. The Jinja is unnecessary here. done.

postgres/client.sls
```diff
@@ -25,23 +22,23 @@ postgresql-client-libs:
       - pkgrepo: postgresql-repo
 {%- endif %}
 
+# Alternatives system. Make client binaries available in $PATH
 {%- if 'bin_dir' in postgres %}
+  {% if postgres.linux.altpriority|int > 0 %}
```
vutny: At first, you don't need to introduce lower level `if` statement, it's better to collapse two statements into one simply using `and` operator. It reduces amount of changes, unnecessary indentations and makes it more readable overall.

At second, no need to use `int` filter here. You could just test with implicit Boolean comparison:
```jinja
  {%- if 'bin_dir' in postgres and
          postgres.linux.altpriority %}
```
This would negate the check if `postgres.linux.altpriority` would be `0`, `''`,`false` or `none`. And that looks better.

postgres/client.sls
```diff
@@ -25,23 +22,23 @@ postgresql-client-libs:
...
-    {%- set path = salt['file.join'](postgres.bin_dir, bin) %}
+    {%- for bin in postgres.client_bins %}
+      {%- set path = salt['file.join'](postgres.bin_dir, bin) %}
 
 {{ bin }}:
   alternatives.install:
     - link: {{ salt['file.join']('/usr/bin', bin) }}
     - path: {{ path }}
-    - priority: 30
+    - priority: {{ postgres.linux.altpriority|int }}
```
vutny: As well, the `int` filter is unnecessary. YAML renderer does not care about Jinja internal data type.

postgres/defaults.yaml
```diff
@@ -41,3 +41,7 @@ postgres:
   databases: {}
   schemas: {}
   extensions: {}
+
+  linux:
+    #Alternatives system are disabled by a 'altpriority=0' pillar.
+    altpriority: 30
```
vutny: Since for now the formula is able to provision the alternatives only for RedHat systems when PG comes from upstream repo, I think the default value across all supported distros must be empty:
```
linux:
  # Alternatives are applicable only for some OS
  altpriority:
```

Please set actual value in `osmap.yaml` file near `bin_dir` and `*_bins` parameters.

postgres/client.sls
noelmcloughlin: Nice suggestion.  This is all done.

PR #180, opened 2018-01-16
Set env=None if cmd.run.env dict is empty

noelmcloughlin:
This PR resolves noisy salt.cmd.run when `cmd.run.env:` is passed empty dictionary. 

` mesg: ttyname failed: Inappropriate ioctl for device`

Context is:
```
[INFO    ] Running state [pg_createcluster 9.5 main] at time 12:32:57.344486
[INFO    ] Executing state cmd.run for [pg_createcluster 9.5 main]
[INFO    ] Executing command 'test -f /var/lib/postgresql/9.5/main/PG_VERSION && test -f /etc/postgresql/9.5/main/postgresql.conf' as user 'root' in directory '/'
mesg: ttyname failed: Inappropriate ioctl for device
[DEBUG   ] output:
[DEBUG   ] Last command return code: [0]
[INFO    ] unless execution succeeded
```

Fix verified on Ubuntu.

### DIFF
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -17,7 +17,7 @@ postgres:
     test: test -f /var/lib/pgsql/data/PG_VERSION
     user: postgres
     # Be aware that there's no ``None`` value exists in YAML
-    env: Null
+    env: []
 
   conf_dir: /var/lib/pgsql/data
   postgresconf: ""
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -16,7 +16,8 @@ postgres:
     command: initdb --pgdata=/var/lib/pgsql/data
     test: test -f /var/lib/pgsql/data/PG_VERSION
     user: postgres
-    env: {}
+    # Be aware that there's no ``None`` value exists in YAML
+    env: Null
 
   conf_dir: /var/lib/pgsql/data
   postgresconf: ""
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -56,7 +56,7 @@ postgresql-cluster-prepared:
     - name: {{ postgres.prepare_cluster.command }}
     - cwd: /
     - runas: {{ postgres.prepare_cluster.user }}
-    - env: {{ postgres.prepare_cluster.env|default({}) or None }}
+    - env: {{ postgres.prepare_cluster.env }}
     - unless:
       - {{ postgres.prepare_cluster.test }}
     - require:
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -56,7 +56,7 @@ postgresql-cluster-prepared:
     - name: {{ postgres.prepare_cluster.command }}
     - cwd: /
     - runas: {{ postgres.prepare_cluster.user }}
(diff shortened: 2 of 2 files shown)


noelmcloughlin:
The `mesg: ttyname failed: Inappropriate ioctl for device` messages comes from salt.utils.vt and there is no easy solution - looks like Ubunut only issue anyway.

postgres/server/init.sls
```diff
@@ -56,7 +56,7 @@ postgresql-cluster-prepared:
     - name: {{ postgres.prepare_cluster.command }}
     - cwd: /
     - runas: {{ postgres.prepare_cluster.user }}
-    - env: {{ postgres.prepare_cluster.env|default({}) }}
+    - env: {{ postgres.prepare_cluster.env|default({}) or None }}
```
vutny: I suggest to edit `defaults.yaml` instead, like this:
```yaml
# defaults.yaml
postgres:
  ...
  prepare_cluster:
    command: initdb --pgdata=/var/lib/pgsql/data
    test: test -f /var/lib/pgsql/data/PG_VERSION
    user: postgres
    env: Null  # Be aware that there's no ``None`` value exists in YAML
```

And just use `- env: {{ postgres.prepare_cluster.env }}` over here.

noelmcloughlin:
Hi @vutny Thanks for your feedbacks.  Funny null did not work either.  Anyway, I have discovered my commit does not fix this issue consistently - there must be some environment issue at play - so rather than travelling down a rabbit hole (its minor issue) 

.. I'll raise a bug on Salt.

My investigation suggest issue arises in `modules/cmdmod.py` or `utils/vt.py`.  I tried setting cmd.run.loglevel: quiet and redirecting cmd.run.test stdin/stdout to /dev/null but that does not help. Some code fails to open a tty terminal and then tries to write to null tty?

noelmcloughlin:
Hey @vutny  Thanks, But to clarify my last message.  This commit does not work - "env: null" is throwing pyyaml error for incorrectly formatted yaml, state fails.  

I can revert the commit or cancel the PR (& raise salt issue instead).

postgres/defaults.yaml
```diff
@@ -16,7 +16,8 @@ postgres:
     command: initdb --pgdata=/var/lib/pgsql/data
     test: test -f /var/lib/pgsql/data/PG_VERSION
     user: postgres
-    env: {}
+    # Be aware that there's no ``None`` value exists in YAML
```
vutny: I think this comment is no longer necessary as well 😃

vutny:
@noelmcloughlin , it looks good. No worries. And you could just amend (`git commit --amend`) your last commit after removing that comment and `push --force` it to this branch. Everything should look pretty.

noelmcloughlin:
@vutny Unfortunately rebasing is disabled on branches off saltstack-formulas master. :-(

`remote: error: GH003: Sorry, force-pushing to cmdrun_env=none_bug is not allowed. `

vutny:
Ah, I thought you're working in your own fork. That's more flexible.

Anyhow, let's close out this PR. Would you mind to submit new one containing only relevant changes?
Thanks.

noelmcloughlin:
Rebased as #183  Thanks!

PR #183, opened 2018-01-22
Env must be list, not dict, in defaults.yml

noelmcloughlin:
Use  '`env: []`' (list) not '`env: {}`' (dict) in defaults.yaml.

(Replaces/rebases #180).


noelmcloughlin:
Thanks @vutny

PR #185, opened 2018-02-05
Support fromrepo in pkg.installed states

### LINKED ISSUE #133: Missing fromrepo on postgresql-server and postgresql-client-libs
While deploying postgres-formula on debian jessie I'm getting a problem. The postgresql-server and postgresql-client-libs don't get installed because of apt pinning. 

```
# apt-cache policy postgresql-common
postgresql-common:
  Installed: (none)
  Candidate: 165+deb8u1
  Version table:
  177.pgdg80+1 0
        500 http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg/main amd64 Packages
     165+deb8u1 0
        990 http://httpredir.debian.org/debian/ jessie/main amd64 Packages
```

You can see 990 for the jessie pkg and 500 for the jessie-pgdg package (less priority).

To get it to work we either have to add some pinning configuration or use fromrepo on the pkg.installed. 

This is the quick hack I did to get it to work (which if necessary should be adapted to work on all distros) : 

```
# git diff
diff --git a/postgres/client.sls b/postgres/client.sls
index c51ca11..6657cb9 100644
--- a/postgres/client.sls
+++ b/postgres/client.sls
@@ -19,6 +19,7 @@ include:
 postgresql-client-libs:
   pkg.installed:
     - pkgs: {{ pkgs }}
+    - fromrepo: jessie-pgdg
 {%- if postgres.use_upstream_repo %}
     - refresh: True
     - require:
diff --git a/postgres/server.sls b/postgres/server.sls
index a553dae..37ad460 100644
--- a/postgres/server.sls
+++ b/postgres/server.sls
@@ -14,6 +14,7 @@ include:
 postgresql-server:
   pkg.installed:
     - pkgs: {{ pkgs }}
+    - fromrepo: jessie-pgdg
 {%- if postgres.use_upstream_repo %}
     - refresh: True
     - require:
```

noelmcloughlin:
This PR fixes a bug around repo handling.  

When `packagename` exists in multiple repo, setting `use_upstream_repo: True` cannot guarantee upstream package gets installed.  This bug was noticed while testing #176 on openSuse with upstream `version: 9.6`.   

Also resolves #133

### DIFF
--- a/postgres/client.sls
+++ b/postgres/client.sls
@@ -19,11 +19,14 @@ include:
 postgresql-client-libs:
   pkg.installed:
     - pkgs: {{ pkgs }}
-{%- if postgres.use_upstream_repo %}
+  {% if postgres.fromrepo %}
+    - fromrepo: {{ postgres.fromrepo }}
+  {% endif %}
+  {%- if postgres.use_upstream_repo == true %}
     - refresh: True
     - require:
       - pkgrepo: postgresql-repo
-{%- endif %}
+  {%- endif %}
 
 {%- if 'bin_dir' in postgres %}
 
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -22,11 +22,11 @@ include:
 postgresql-server:
   pkg.installed:
     - pkgs: {{ pkgs }}
-{%- if postgres.use_upstream_repo %}
+  {%- if postgres.use_upstream_repo == true %}
     - refresh: True
     - require:
       - pkgrepo: postgresql-repo
-{%- endif %}
+  {%- endif %}
 
 {%- if 'bin_dir' in postgres %}
 
--- a/postgres/codenamemap.yaml
+++ b/postgres/codenamemap.yaml
@@ -22,6 +22,7 @@
 
 {{ codename|default(name, true) }}:
   # PostgreSQL packages are mostly downloaded from `main` repo component
+  fromrepo: {{ name }}
   pkg_repo:
     name: 'deb http://apt.postgresql.org/pub/repos/apt {{ name }}-pgdg main {{ repo.version }}'
   pkg: postgresql-{{ version }}
@@ -55,6 +56,7 @@
 
 {{ codename|default(name, true) }}:
   # PostgreSQL packages are mostly downloaded from `main` repo component
+  fromrepo: {{ name }}
   pkg_repo:
     baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ version }}/fedora/fedora-$releasever-$basearch'
 
@@ -81,6 +83,9 @@
 
 ## Fedora
 # `oscodename` grain has long distro name
+{{ fedora_codename('Fedora-28', '10.3', 'Fedora 28 (Twenty Eight)') }}
+{{ fedora_codename('Fedora-27', '9.6', 'Fedora 27 (Twenty Seven)') }}
+{{ fedora_codename('Fedora-26', '9.6', 'Fedora 26 (Twenty Six)') }}
 {{ fedora_codename('Fedora-25', '9.5', 'Fedora 25 (Twenty Five)') }}
 {{ fedora_codename('Fedora-24', '9.5', 'Fedora 24 (Twenty Four)') }}
 {{ fedora_codename('Fedora-23', '9.4', 'Fedora 23 (Twenty Three)') }}
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -36,6 +36,10 @@ postgres:
 
   bake_image: False
 
+  fromrepo:
+  pkg_repo:
+    name: []
+
   users: {}
   tablespaces: {}
   databases: {}
--- a/postgres/dev.sls
+++ b/postgres/dev.sls
@@ -1,13 +1,19 @@
 {% from "postgres/map.jinja" import postgres with context %}
 
-{% if postgres.pkg_dev %}
+  {% if postgres.pkg_dev %}
 install-postgres-dev-package:
   pkg.installed:
     - name: {{ postgres.pkg_dev }}
-{% endif %}
+    {% if postgres.fromrepo %}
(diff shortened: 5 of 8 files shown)


noelmcloughlin:
Still testing this.

noelmcloughlin:
Testing done - works nicely.

postgres/server/init.sls
```diff
@@ -22,11 +16,16 @@ include:
...
+  {%- if postgres.pkg_repo.name %}
+    - fromrepo: {{ postgres.pkg_repo.name }}
+  {%- endif %}
+  {%- if postgres.use_upstream_repo %}
     - refresh: True
     - require:
       - pkgrepo: postgresql-repo
-{%- endif %}
+
+    {%- do includes.append('postgres.upstream') %}
```
vutny: We have already substituted the `includes` list above, so this line has no effect and the `postgres.upstream` states would never be included. It is always better to keep definition and logic around `includes` statements at the top of SLS file. That looks simpler and makes debugging easier. By removing a single `if ` statement you don't make the code more readable or "optimized".

postgres/client.sls
```diff
@@ -7,23 +7,23 @@
...
-{%- endif %}
-
 # Install PostgreSQL client and libraries
 
 postgresql-client-libs:
   pkg.installed:
     - pkgs: {{ pkgs }}
-{%- if postgres.use_upstream_repo %}
+  {%- if postgres.pkg_repo.name %}
+    - fromrepo: {{ postgres.pkg_repo.name }}
```
vutny: This kind of the `fromrepo`  usage would work for YUM etc, but not for APT. Please check out the [docs](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkgrepo.html#salt.states.pkgrepo.managed).
You need to provide distribution codename for Debian-based distros. Better to do it by exposing parameter `dist` in `codenamemap.yaml` file and use `pkg_repo.name` only as a fallback.

postgres/server/init.sls
noelmcloughlin: In support of opinions in #125, its good idea to removing some Jinja in SLS/YML files. These sentiments, which I tend to agree with, can only be addressed in baby steps. No issue here.

I moved this includes down few lines in the code - its the first occurance so surely takes effect.

postgres/client.sls
noelmcloughlin: Thanks, this is good feedback. I have only tested on YUM system, so good catch. I appreciate your PR reviews @vutny, they are always sharp and helpful to community!!  I'll do more testing and see if this blocks the PR.

postgres/server/init.sls
vutny: I saw you did it differently in the `client.sls`. In THAT case it would work, despite not looking pretty though...

Please double check: moving `do includes.append...` statement here would break calling the single state like `salt-call state.apply postgres.server`. The `postgres.upstream` SLS would be never included this way.

noelmcloughlin: The server SLS is more tricky due to `includes` Jinja.  And reviewing this, maybe the following Jinja must move below  `postgresql.server.pkg.installed` state so indent(2) is applied?
```
{%- if includes -%}
include:
  {{ includes|yaml(false)|indent(2) }}
{%- endif %}
```
The usage of `salt.apply` is new for me. I'll will report back.

noelmcloughlin:
The latest commit works on Ubuntu, Fedora, Centos7 for downstream. 
Needs testing on upstream, maybe opensuse too.

**Ubuntu**
```
          ID: postgresql-server
    Function: pkg.installed
      Result: True
     Comment: 2 targeted packages were installed/updated.
     Started: 09:45:18.391633
    Duration: 19893.976 ms
     Changes:   
              ----------
              postgresql-9.5:
                  ----------
                  new:
                      9.5.11-0ubuntu0.16.04
                  old:
              postgresql-client:
                  ----------
                  new:
                      1
                  old:
              postgresql-client-9.5:
                  ----------
                  new:
                      9.5.11-0ubuntu0.16.04
                  old:
              postgresql-client-common:
                  ----------
                  new:
                      173ubuntu0.1
                  old:
              postgresql-common:
                  ----------
                  new:
                      173ubuntu0.1
                  old:
              postgresql-contrib:
                  ----------
                  new:
                      9.5+173ubuntu0.1
                  old:
              postgresql-contrib-9.5:
                  ----------
                  new:
                      9.5.11-0ubuntu0.16.04
                  old:
----------
    Function: cmd.run
        Name: pg_createcluster 9.5 main -d  /var/lib/postgresql/9.5/main
      Result: True
     Comment: Command "pg_createcluster 9.5 main -d  /var/lib/postgresql/9.5/main" run
     Started: 09:45:38.295506
    Duration: 5597.467 ms
     Changes:   
              ----------
              pid:
                  34997
              retcode:
                  0
              stderr:
              stdout:
                  Creating new cluster 9.5/main ...
                    config /etc/postgresql/9.5/main
                    data   /var/lib/postgresql/9.5/main
                    locale C
                    socket /var/run/postgresql
                    port   5432
```

vutny:
The `use_upstream_repo: False` setting will cause trouble here because it should trigger inclusion of `postgres.upstream` state.

noelmcloughlin:
The only ` use_upstream_repo: False` setting I can find is in `pillar.example` and PR does not change.  So far I have tested the PR with distro supplied Postgres and `False` is needed there.

For next test, using upstream, I plan to set `use_upstream_rep: True` pillar.  Is this what you are reminding me about?

vutny:
Oh yes! Sure thing, I meant `use_upstream_repo: True` of course. :sleepy:

noelmcloughlin:
Testing this today.

noelmcloughlin:
Closing to rebase & submit smaller PRs.  Comments here will be adopted.

noelmcloughlin:
New [commit done](https://github.com/saltstack-formulas/postgres-formula/commit/a4e184b79cebf8018803d657c622939654f76653), PR later.

PR #182, opened 2018-01-17
Add basic postgres.server.remove state

noelmcloughlin:
This adds new `postgres.server.remove, postgres.client.remove` states to the formula.   Please review and provide feedback or merge if approved. Thanks.    

The DB directory (i.e.` postgres.conf_dir`) is not explicitly removed by the formula.

**Upstream remove**
```
          ID: postgresql-repo-removed
    Function: pkgrepo.absent
        Name: deb http://apt.postgresql.org/pub/repos/apt xenial-pgdg main 9.6
      Result: True
     Comment: Removed repo deb http://apt.postgresql.org/pub/repos/apt xenial-pgdg main 9.6, and keyid ACCC4CF8
     Started: 15:48:23.883788
    Duration: 5301.947 ms
     Changes:   
              ----------
              keyid:
                  ACCC4CF8
              repo:
                  deb http://apt.postgresql.org/pub/repos/apt xenial-pgdg main 9.6
----------
          ID: postgresql-notrunning
    Function: service.dead
        Name: postgresql
      Result: True
     Comment: The service postgresql is already dead
     Started: 15:48:30.023758
    Duration: 69.957 ms
     Changes:   
----------
          ID: postgresql-server-removed
    Function: pkg.removed
      Result: True
     Comment: All targeted packages were removed.
     Started: 15:48:30.495447
    Duration: 5205.095 ms
     Changes:   
              ----------
              postgresql-9.6:
                  ----------
                  new:
                  old:
                      9.6.6-1.pgdg16.04+1
              postgresql-contrib-9.6:
                  ----------
                  new:
                  old:
                      9.6.6-1.pgdg16.04+1
              postgresql-plpython:
                  ----------
                  new:
                  old:
                      1
              postgresql-plpython-9.6:
                  ----------
                  new:
                  old:
                      9.6.6-1.pgdg16.04+1
----------
          ID: postgresql-config-removed
    Function: file.absent
        Name: /etc/postgresql/9.6/main
      Result: True
     Comment: File /etc/postgresql/9.6/main is not present
     Started: 15:48:35.712664
    Duration: 1.392 ms
     Changes:   
----------
          ID: postgresql-tablespace-dir-my_space-removed
    Function: file.absent
        Name: /srv/my_tablespace
      Result: True
     Comment: Removed directory /srv/my_tablespace
     Started: 15:48:35.715247
    Duration: 29.948 ms
     Changes:   
              ----------
              removed:
                  /srv/my_tablespace

Summary for local
------------
[...]

PR #187, opened 2018-02-10
allow for separate configuration and data directories

noelmcloughlin:
This PR allows different "configuration" and "data" directory, resolving some issues-

1. Ubuntu postgresql-server stores config in `/etc`, and data into `/var`.  The `postgres` state fails horribly if `/etc/postgresql/../postgresql.conf` exists, perhaps after `postgres.dropped` runs. 

2. User may want separation of data and configuration.

3. This is **current situation** where `conf_dir` variable exists but not `data_dir`.
```
codenamemap.yaml:  conf_dir: /etc/postgresql/{{ version }}/main
defaults.yaml:  conf_dir: /var/lib/pgsql/data
osmap.yaml:  conf_dir: /var/lib/postgres/data
osmap.yaml:  conf_dir: /var/lib/pgsql/{{ repo.version }}/data
osmap.yaml:  conf_dir: /usr/local/var/postgres
```

4. `postgres` install always fails after 'dropped' (or remove #182) states, because conf_dir remains.

5. If `conf_dir` and `data_dir` have same value, `pg_ctl` throws permissions error.

`pg_ctl[8774]: FATAL:  data directory "/var/lib/pgsql/data" has group or world access`

### DIFF
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -52,21 +52,41 @@ postgresql-server:
 {%- endif %}
 
 postgresql-cluster-prepared:
+  file.directory:
+    # The next state runs as unprivledged user but "unless" check needs privledge.
+    # This state ensures "cmd.run.unless='test -f data_dir/PG_VERSION'" works below.
+    - name: {{ postgres.data_dir }}
+    - dir_mode: 0755
+    - onlyif: test -f {{ postgres.data_dir }}/{{ postgres.prepare_cluster.pgtestfile }}
   cmd.run:
+ {%- if postgres.prepare_cluster.command is defined %}
+      {# support for depreciated 'prepare_cluster.command' pillar #}
     - name: {{ postgres.prepare_cluster.command }}
+    - unless: {{ postgres.prepare_cluster.test }}
+ {%- else %}
+    - name: {{ postgres.prepare_cluster.pgcommand + ' ' }} {{ postgres.data_dir }}
+    - unless: test -f {{ postgres.data_dir }}/{{ postgres.prepare_cluster.pgtestfile }}
+ {%- endif %}
     - cwd: /
     - runas: {{ postgres.prepare_cluster.user }}
     - env: {{ postgres.prepare_cluster.env }}
-    - unless:
-      - {{ postgres.prepare_cluster.test }}
     - require:
+      - file: postgresql-cluster-prepared
       - pkg: postgresql-server
 
 postgresql-config-dir:
   file.directory:
-    - name: {{ postgres.conf_dir }}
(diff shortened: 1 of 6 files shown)


postgres/server/init.sls
```diff
@@ -70,6 +70,9 @@ postgresql-config-dir:
     - makedirs: True
     - require:
       - cmd: postgresql-cluster-prepared
+    {%- if postgres.conf_dir == postgres.data_dir %}
+    - dir_mode: 0700
```
vutny: This is a good change, and the configuration directory should definitely have to be secured even if does not match the data directory. No need for `if` statement. Just put it above of the `require` items as a matter of consistent style.

noelmcloughlin: I was reading PG documentation regarding this (due to error in logs).  I could not find any instruction that conf_dir needs to be secured.  I understand you see no issues using 0700 for conf_dir so agree here.

noelmcloughlin:
The primary PR goal is to fix 0700 permission bug.  PG documentation/error complains about _data_dir_ not configuration directory.

In _default.yaml_ the '/var/lib/pgsql/data' string occurs 3 times to avoid Jinja, and be consistent.

In _osmap.yaml_ the _data_dir_  is declared as Jinja, not YAML. I'm fixing the inconsistency cautiously. It is better for everyone to address the YAML first, and SLS in later PR.  Researching correct values for osmap.yaml involved significant research into various wiki's and distro docs.  Can we just ensure we are happy with YAML first.

postgres/server/init.sls
noelmcloughlin: I have implemented this. Because it's allowed to have cond_dir and data_dir the same in PG implementations, and 0700 was already requirement, then 0700 applies to conf_dir too.  You are correct therefore - the if should be removed.

noelmcloughlin:
In hindsight, its better to do SLS now.  Implementation wise, we could remove `command` and `test` from this block, and move command/test definition to SLS.
```
prepare_cluster
    command: initdb --pgdata=/var/lib/pgsql/data
    test: test -f /var/lib/pgsql/data/PG_VERSION
```
Perhaps this is best patten so PG command is not hardcoded. But this implies '--pgdata=' should be hardcoded in SLS.
```
prepare_cluster
    command: initdb
```
@vutny What do you suggest?

noelmcloughlin:
@vutny Please review this again when you have free time. thx

noelmcloughlin:
Testing this today.

noelmcloughlin:
Closing to rebase & submit smaller PRs.  Comments here will be adopted.

PR #192, opened 2018-03-17
Fix suse alternatives #189

### LINKED ISSUE #189: Is bin_dir and *bins missing for Suse in osmap.yml
Check if `osmap.yaml` requires `bin_dir `and `*_bins` parameter are needed, similar to RedHat section.  If they are needed add this if #176 is merged.  The packaging maintainers may handle this (if memory serves correctly).

noelmcloughlin:
Resolves #189


aboe76:
@noelmcloughlin thanks

PR #193, opened 2018-03-20
corrected package defaults/maps

noelmcloughlin:
PR to tidyup few minor details in defaults / osmap.


noelmcloughlin:
Thanks @vutny    Verified on Fedora 27 twice - upstream=true, upstream=false.

PR #200, opened 2018-03-28
Avoid corrupted upstream install on Debian's

noelmcloughlin:
When using **upstream repo** on Debian family, the formula installs both upstream and downstream PG releases to satisfy `postgres-server-dev-all` dependencies. This is unwanted situation.

This PR avoids corrupted upstream PG installation on Debian family.


noelmcloughlin:
Thanks @johnkeates

johnkeates:
I love this type of concise and targeted PR's, very easy to fast-track and helps a ton.

PR #196, opened 2018-03-20
Fedora repo

noelmcloughlin:
Use [Fedora upstream repo](https://download.postgresql.org/pub/repos/yum/10/fedora/). Verified on Fedora 27.

PR #201, opened 2018-03-28
Use osmap for OS handling of fedora repo

noelmcloughlin:
This PR is better implementation of #196 (Fedora repo). The OS specific code for Fedora is moved to osmap.yaml which is cleaner and removes some Jinja code.

### DIFF
--- a/postgres/osmap.yaml
+++ b/postgres/osmap.yaml
@@ -1,0 +1,7 @@
+{% import_yaml "postgres/repo.yaml" as repo %}
+
+Fedora:
+  pkg_repo:
+    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/fedora/fedora-$releasever-$basearch'
+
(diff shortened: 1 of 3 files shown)


postgres/osmap.yaml
```diff
@@ -0,0 +1,7 @@
+{% import_yaml "postgres/repo.yaml" as repo %}
+
+Fedora:
+  pkg_repo:
+    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/fedora/fedora-$releasever-$basearch'
+
+# vim: ft=yaml
```
vutny: Side comment. If you use (or would like to use) [salt-vim](https://github.com/saltstack/salt-vim) plugin, it's better to set file type to sls here: `vim: ft=sls`. It will enable correct Jijna syntax highlighting.

noelmcloughlin:
Thanks @vutny I have adopted your comment.

PR #204, opened 2018-03-31
use upstream repo for python if configured

noelmcloughlin:
For some reason `python.sls` was missing upstream repo support. PR to fix that.


noelmcloughlin:
Thanks for taking time to review @vutny.  Rebased.  I implemented your review comments.

However, `server/init.sls` was my reference point for need to conditionally `include: - upstream.sls`. But since we agree a conditional include of `upstream.sls` is superfluous, a future PR should address this in `server/init.sls` too - just to be consistent across SLS files.

**server/init.sls**
```
{%- set includes = [] %}
... 
{%- if postgres.use_upstream_repo == true -%}
  {%- do includes.append('postgres.upstream') %}
{%- endif %}
```

PR #207, opened 2018-04-12
Fix compatibility with Salt 2018.3.0

vutny:
This PR makes sure that data structures like nested lists (or dictionaries) are properly translated to valid YAML syntax. It also makes the formula work with latest stable version of Salt: 2018.3.0.


javierbertoli:
Great!

vutny:
Thank you @javierbertoli !

PR #210, opened 2018-05-03
Better default data directory permission

mkotsbak:
Postgresql won't start without directory permission 0700.

### DIFF
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -21,7 +21,7 @@ postgres:
     env: []
 
   conf_dir: /var/lib/pgsql/data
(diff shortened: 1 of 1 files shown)


noelmcloughlin:
Maybe already solved in https://github.com/saltstack-formulas/postgres-formula/pull/202 - instead the README & pillar.example may need update.

mkotsbak:
Not sure if the current default let Postgres start before? Else it does not make much sense to have that default. I was able to fix it by specifying another permission yes.

noelmcloughlin:
Thanks @mkotsbak and @vutny

PR #212, opened 2018-05-09
Tidyup codenamemap.yaml for Fedora

noelmcloughlin:
This PR is small tidyup following PR #209 (add fromrepo support).

1) The Fedora section is missing `{% set fromrepo ....}` statements - my mistake.
2) Repeating `{% if repo.use_upstream_repo == true %}` block is unnecessary, use one only.

### DIFF
--- a/postgres/codenamemap.yaml
+++ b/postgres/codenamemap.yaml
@@ -2,18 +2,13 @@
 
 {% import_yaml "postgres/repo.yaml" as repo %}
 
-{% macro debian_codename(name, version, codename=none) %}
-  {#
-  Generate lookup dictionary map for Debian and derivative distributions
-
-    name:
-      distro codename
-    version:
-      PostgreSQL release version
-    codename:
-      optional grain value if `name` does not match the one returned by
-      `oscodename` grain
+  {# Generate lookup dictionary map for OS and derivative distributions
+    name: distro codename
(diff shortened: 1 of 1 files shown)


postgres/codenamemap.yaml
```diff
@@ -40,22 +40,6 @@
...
-      PostgreSQL release version
-    codename:
-      optional grain value if `name` does not match the one returned by
-      `oscodename` grain
-  #}
-
-  {# use upstream version if configured #}
-  {% if repo.use_upstream_repo == true %}
-    {% set version = repo.version %}
-  {% endif %}
```
vutny: You have moved `version` variable re-assignment out of `macro` scope, so it would not be altered then you have to set version for upstream packages. All macros calls will use only the version value given by the parameter, ignoring the Pillar-given value.

noelmcloughlin:
Good spot @vutny!!  I've reverted back since `version` and `name` are both macro parameters. 
 Instead the PR fixes the missing `fromrepo` for Fedora distributions.  Thanks.

aboe76:
@vutny can you take another look please?

aboe76:
@noelmcloughlin andd @vutny thanks for your work and time.

noelmcloughlin:
Thanks @aboe76 @vutny

PR #215, opened 2018-06-05
Update `codenamemap` for Ubuntu

myii:
Needed for `bionic`.  Modified the whole list according to the current supported versions as shown at:

* https://packages.ubuntu.com/search?keywords=postgresql&searchon=names


myii:
You're welcome.

Issue #214, opened 2018-05-22
State postgres-reload-modules always reports changes


Magomogo:
When a state does nothing it shouldn't report changes.
The issue caused by https://github.com/saltstack-formulas/postgres-formula/pull/188

PR #216, opened 2018-06-05
Makes postgres-reload-modules conditional

RobRuana:
This pull request makes the `postgres-reload-modules` conditional so it doesn't report changes during every run.

### DIFF
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -57,6 +57,8 @@ postgres:
 
   bake_image: False
 
+  manage_force_reload_modules: True
+
   fromrepo:
 
   users: {}
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -57,7 +57,7 @@ postgres:
 
   bake_image: False
 
-  manage_force_reload_modules: True
+  manage_force_reload_modules: False
 
   fromrepo:
 
--- a/postgres/manage.sls
+++ b/postgres/manage.sls
@@ -1,7 +1,9 @@
 {%- from tpldir + "/map.jinja" import postgres with context -%}
 {%- from tpldir + "/macros.jinja" import format_state with context -%}
 
-{%- if salt['postgres.user_create']|default(none) is not callable %}
+{%- set needs_client_binaries = salt['postgres.user_create']|default(none) is not callable -%}
+
+{%- if needs_client_binaries %}
 
 # Salt states for managing PostgreSQL is not available,
 # need to provision client binaries first
@@ -15,11 +17,15 @@ include:
 
 {%- endif %}
 
+{%- if needs_client_binaries or postgres.manage_force_reload_modules %}
+
 # Ensure that Salt is able to use postgres modules
 
 postgres-reload-modules:
   test.succeed_with_changes:
     - reload_modules: True
+
+{%- endif %}
 
 # User states
 
--- a/postgres/manage.sls
+++ b/postgres/manage.sls
@@ -17,7 +17,6 @@ include:
 
 {%- endif %}
 
-
 # Ensure that Salt is able to use postgres modules
 
 postgres-reload-modules:
--- a/postgres/manage.sls
+++ b/postgres/manage.sls
@@ -17,15 +17,16 @@ include:
 
 {%- endif %}
 
-{%- if needs_client_binaries or postgres.manage_force_reload_modules %}
 
 # Ensure that Salt is able to use postgres modules
 
 postgres-reload-modules:
+{%- if needs_client_binaries or postgres.manage_force_reload_modules %}
   test.succeed_with_changes:
-    - reload_modules: True
-
+{%- else %}
+  test.succeed_without_changes:
 {%- endif %}
+    - reload_modules: True
 
 # User states
 
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -101,8 +101,6 @@ postgresql-conf:
     - backup: {{ postgres.config_backup|default(false, true) }}
     - require:
       - file: postgresql-config-dir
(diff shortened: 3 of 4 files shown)


postgres/codenamemap.yaml
```diff
@@ -70,6 +70,7 @@
 {{ debian_codename('vivid', '9.4') }}
 {{ debian_codename('wily', '9.4') }}
 {{ debian_codename('xenial', '9.5') }}
+{{ debian_codename('bionic', '10') }}
```
RobRuana: Adds codename for Ubuntu 18.04 "bionic". This was all I needed to change to get Postgres 10 working for Ubuntu 18.04.

postgres/defaults.yaml
```diff
@@ -57,6 +57,8 @@ postgres:
 
   bake_image: False
 
+  manage_force_reload_modules: True
```
RobRuana: New config setting defaults to `True` to preserve existing behavior.

postgres/manage.sls
```diff
@@ -15,12 +17,16 @@ include:
 
 {%- endif %}
 
+{%- if needs_client_binaries or postgres.manage_force_reload_modules %}
```
RobRuana: I was getting tired of `postgres-reload-modules` running even when nothing had changed. I hid the state behind a conditional that checks if any new client binaries were installed, or if `manage_force_reload_modules` was set to `True`.

*NOTE*: `manage_force_reload_modules` defaults to `True` to preserve the existing behavior.

myii:
@RobRuana Great minds think alike... pipped you to the `postgres/codenamemap.yaml` change by 3 hours -- see #215.  Took it a bit further by updating the list to all of the supported versions and getting rid off the deadwood.

postgres/manage.sls
noelmcloughlin: Maybe`mange_force_reload_modules` should default to False to fix  https://github.com/saltstack-formulas/postgres-formula/issues/214  @vunty what do you think?

vutny: I'm totally fine with setting `mange_force_reload_modules` to `False` by default.

RobRuana:
Okay, I merged the codename changes from `master` and made `manage_force_reload_modules` default to `False`.

Furthermore, I repositioned the conditional so `postgres-reload-modules` always runs, so other states will always be able to reference it. The part that's conditional is whether it reports changes or not.

postgres/manage.sls
```diff
@@ -18,7 +20,11 @@ include:
 # Ensure that Salt is able to use postgres modules
 
 postgres-reload-modules:
+{%- if needs_client_binaries or postgres.manage_force_reload_modules %}
   test.succeed_with_changes:
+{%- else %}
+  test.succeed_without_changes:
+{%- endif %}
```
vutny: @RobRuana This seems like an obvious fix, but that wouldn't work.
Each state generated below explicitly subscribed on changes in `postgres-reload-modules`. So if you later change PG entities in the Pillar, it will not have any effect.
The only way I see to overcome this: you need to put conditional in `format_state` macro to depend on modules reload only when `postgres.manage_force_reload_modules` set `True`.

postgres/server/init.sls
```diff
@@ -184,9 +184,6 @@ postgresql-running:
   service.running:
     - name: {{ postgres.service }}
     - enable: True
-   {% if grains.os not in ('MacOS',) %}
-    - reload: True
-   {% endif %}
```
myii: @RobRuana This has already been resolved by 79ab1a2 during #218:

https://github.com/saltstack-formulas/postgres-formula/blob/332fe7b2b57f8a1096e91f5920acdea9fd2192e8/postgres/server/init.sls#L141-L151

RobRuana: Thanks for the notice! I'm scrambling to get everything working at the moment, and this was my quick and dirty fix. I've synced with upstream.

myii: @RobRuana You're welcome.

aboe76:
@vutny could you take another look and ping me if it's ok?

vutny:
@aboe76 @RobRuana My in-line comment above still applies. All latter states in `postgres.manage` are subscribed on changes from `postgres-reload-modules`. So if it succeeds without changes in subsequent highstate run, there would be no further states to trigger, even if related Pillar data would be changed.

postgres/manage.sls
RobRuana: @vutny, yes, you are correct that this breaks any state using the `format_state()` macro.

At first I couldn't understand why `onchanges: postgres-reload-modules` was being used instead of `require: postgres-reload-modules`. But then I realized that many instances of `format_state()` are already using the `require` conditional, so there would be conflicting `require` keys when the state is rendered.

Correct me if I'm wrong, but I believe `onchanges` is being used to force correct ordering of the states, not because we want every `format_state()` to run when there are changes in `postgres-reload-modules`.

If that's correct, then we should be able to achieve the same result by using `watch` instead of `onchanges` in `format_state()`. I've made that change. Please let me know if you think that works.

postgres/server/init.sls
```diff
@@ -176,6 +169,15 @@ postgresql-pg_hba:
 {%- endif %}
     - require:
       - file: postgresql-config-dir
+    - watch_in:
+      - module: postgresql-service-restart
```
RobRuana: I've found that simply reloading the postgres service is not enough to pick up ACL changes in pg_hba.conf. A full restart is required.

This is on ubuntu 18.04 with postgres 10.

RobRuana:
I'm going to close this pull request for now. I may revisit this issue, but for now I am okay with the `postgres-reload-modules` reporting changes all the time.

Issue #174, opened 2017-10-06
Add support for multiple clusters / versions


PhilippeAB:
Need the port to be configurable

myii:
@PhilippeAB Port is now configurable.

myii:
@noelmcloughlin I don't think this issue should be closed just yet.  Allowing the port to be configurable was just the first stage.  As I've said in https://github.com/saltstack-formulas/postgres-formula/pull/218#issuecomment-397809521, this progress can now be built upon to support multiple clusters/versions.

noelmcloughlin:
True. I closed because the issue description only requested a configurable port. I now see slogan is talking about multiple cluster / version. Thats fine.

Issue #43, opened 2015-03-11
Convert the cmd.run for creating clusters to using the cluster execution module


arthurlogilab:
https://github.com/saltstack/salt/pull/21314 has been merged into develop, it could be used instead of doing cmd.run 

States should be arriving shortly.

iggy:
That won't make it into a release for quite some time (2015.9?). Even then most people aren't going to immediately move to the latest version. So... long term goal I guess.

arthurlogilab:
We have ended up using clusters in a use case of some postgreql servers deployed using this formula. 

Right now the formula works in most places by adding db_port when needed (in some cases we need to use the conf_dir), but we might have to develop a few things on the formula that we will contribute if we find the time to adapt the formula.

myii:
@arthurlogilab Any plans for implementing this now?

arthurlogilab:
Sorry no.

PR #218, opened 2018-06-08
Allow `port` to be configurable

myii:
#### References

1. #174 
1. https://github.com/saltstack-formulas/postgres-formula/issues/43#issuecomment-302681240

#### Discussion

I've been using this formula for a while now and like the comment linked above, I've been able to use `db_port` to adapt to clusters not using the default port of `5432`.

The implementation does not affect existing configurations.  The non-standard `port` would have to be explicitly provided in the pillar.

Changing the port requires a service restart, so the existing service reload state doesn't work -- hence using `service.running` again.  More could be done to check the various situations but I preferred simplicity to efficiency, at least when introducing this functionality.

I envisage that this is a useful first step towards adding support for multiple clusters / versions (i.e. #174).

#### Concerns

The will probably necessitate modification of:

https://github.com/saltstack-formulas/postgres-formula/blob/800259dffe5b89f11d4198d2bd5df163bad5d9aa/test/integration/default/serverspec/postgres_spec.rb#L10

#### Regex description

```regex
- `^`        Line start
- `#*\s*`    Find line even if commented out
- `(port)`   'port' -- capture as backreference `\1`
- `\s*=\s*`  Equals sign, whether or not surrounded by whitespace
- `\d{4,5}`  Existing port value, expected at 4/5 digits
- `(.*)`     Remainder (i.e. comment) -- capture as backreference `\2`
- `$`        Line end
```

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -32,6 +32,10 @@ postgres:
     hard: 64000
 
   # POSTGRES
+  # Non-standard port to use for the cluster
+  # Only set if port `5432` is not appropriate
+  port: 5433
+
   # Append the lines under this item to your postgresql.conf file.
   # Pay attention to indent exactly with 4 spaces for all lines.
   postgresconf: |
--- a/postgres/macros.jinja
+++ b/postgres/macros.jinja
@@ -23,6 +23,7 @@
 {{ state }}-{{ name }}:
   {{ state }}.{{ ensure|default('present') }}:
     {{- format_kwargs(kwarg) }}
+    - db_port: {{ postgres.port|default(postgres.default_port) }}
     - onchanges:
       - test: postgres-reload-modules
 
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -192,3 +192,24 @@ postgresql-running:
       - file: postgresql-pg_ident
 
 {%- endif %}
+
+{%- if postgres.port|default(false) %}
+
+postgresql-port:
+  file.replace:
+    - name: {{ postgres.conf_dir }}/postgresql.conf
+    - pattern: ^#*\s*(port)\s*=\s*\d{4,5}(.*)$
+    - repl: \1 = {{ postgres.port }}\2
+    - flags: 8  # ['MULTILINE']
+    - show_changes: True
+    - append_if_not_found: True
+    - backup: {{ postgres.config_backup|default(false, true) }}
+    - require:
+      - file: postgresql-config-dir
+    - watch_in:
+       - service: postgresql-port
+
+  service.running:
+    - name: {{ postgres.service }}
+
+{%- endif %}
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -3,6 +3,7 @@
 postgres:
   use_upstream_repo: True
   version: '9.5'
+  default_port: 5432
   pkg: postgresql
   pkgs_extra: []
   pkg_client: postgresql-client


myii:
#### Current deficiencies

One way this method could be disadvantageous is that there is no comment/warning that the port is being managed by Salt.  Couple of suggestions how this could be resolved:

1. Simply add a comment above the `port` setting to warn that it is being managed by Salt
    1. Would have to be careful that this doesn't keep getting set which each subsequent run
    1. Probably safer to just append the comment on the same line, instead of backreference `\2`
1. Another solution could be to disable the `port` in the main body and then append the setting using a `file.blockreplace`, like is being done for `listen_addresses`

myii:
#### Comment appended to same line

Used the simplest solution to append the warning to the same line:

```diff
--- 
+++ 
@@ -60,7 +60,7 @@
                                      # comma-separated list of addresses;
                                      # defaults to 'localhost'; use '*' for all
                                      # (change requires restart)
-port = 5433                          # (change requires restart)
+port = 5433                          # Managed by SaltStack: please do not edit
 max_connections = 100                # (change requires restart)
 #superuser_reserved_connections = 3  # (change requires restart)
 unix_socket_directories = '/var/run/postgresql'      # comma-separated list of directories
```

pillar.example
```diff
@@ -32,6 +32,10 @@ postgres:
     hard: 64000
 
   # POSTGRES
+  # Non-standard port to use for the cluster
+  # Only set if port `5432` is not appropriate
+  port: 5433
+
```
vutny: With what I said above, we could provide top level `postgres.port` Pillar setting. This would be reused directly in the server SLS and automatically enable correct port resolution for all Salt routines related to PostgreSQL.

But just better to keep it default mentioning it could be changed.

postgres/macros.jinja
```diff
@@ -23,6 +23,7 @@
 {{ state }}-{{ name }}:
   {{ state }}.{{ ensure|default('present') }}:
     {{- format_kwargs(kwarg) }}
+    - db_port: {{ postgres.port|default(postgres.default_port) }}
```
vutny: With setting `postgres.port` this would be unnecessary.

postgres/server/init.sls
```diff
@@ -192,3 +192,25 @@ postgresql-running:
...
+    - show_changes: True
+    - append_if_not_found: True
+    - backup: {{ postgres.config_backup|default(false, true) }}
+    - require:
+      - file: postgresql-config-dir
+    - watch_in:
+       - service: postgresql-port
+
+  service.running:
+    - name: {{ postgres.service }}
```
vutny: I think having two states handling a single service is not that good idea, since this could be very easy overlooked in further development...

But I understand we'd have to deal with `port` setting could be defined outside of `Managed by SaltStack` protected zone. So we can add a `file.comment` state with rather simple regex to force commenting out string matching `/port = .+/` **only when** `postgres.port` is defined and there would be changes in `postgresql-conf`:

```yaml
{%- if salt['pillar.get']('postgres.port') %}

postgresql-alter-port:
  file.comment:
    - name: {{ postgres.conf_dir }}/postgresql.conf
    - regex: ^port = .+
    - prereq:
      - file: postgresql-conf

{%- endif %}
```

myii:
@vutny Thanks for your feedback.  I'll get back to you once I get the chance to go over the points you've raised.

myii:
@vutny Apologies for the delay, needed to find some free time to work on this.  Changes should be clear enough, since mainly an incorporation of your suggestions.

pillar.example
```diff
@@ -34,7 +38,7 @@ postgres:
   # POSTGRES
   # Append the lines under this item to your postgresql.conf file.
   # Pay attention to indent exactly with 4 spaces for all lines.
-  postgresconf: |
+  postgresconf: |-
```
myii: Otherwise an unnecessary blank line rendered, more noticeable if both `listen_addresses` and `port` are given.

postgres/server/init.sls
```diff
@@ -86,23 +86,43 @@ postgresql-config-dir:
     - require:
       - cmd: postgresql-cluster-prepared
 
-{%- if postgres.postgresconf %}
+{%- set db_port = salt['config.option']('postgres.port') %}
```
myii: Instead of the suggested:

```sls
{%- if salt['pillar.get']('postgres.port') %}
```

* `config.option` gets the value from the minion configuration first, otherwise the pillar (which then works correctly with the `postgres.manage` states)
* Same as the method used in the code for `salt.modules.postgres`

postgres/server/init.sls
```diff
@@ -86,23 +86,43 @@ postgresql-config-dir:
...
+      - file: postgresql-config-dir
+
+{%- endif %}
+
+{%- if postgres.postgresconf or db_port %}
 
 postgresql-conf:
   file.blockreplace:
     - name: {{ postgres.conf_dir }}/postgresql.conf
     - marker_start: "# Managed by SaltStack: listen_addresses: please do not edit"
```
myii: Obviously can't change this marker to mention that the managed block is also for the `port`.

postgres/server/init.sls
```diff
@@ -86,23 +86,43 @@ postgresql-config-dir:
     - require:
       - cmd: postgresql-cluster-prepared
 
-{%- if postgres.postgresconf %}
+{%- set db_port = salt['config.option']('postgres.port') %}
+{%- if db_port %}
+
+postgresql-conf-comment-port:
```
myii: > Since the `port` setting is commented in the PG distribution, this is rather safe.

Unfortunately, the `port` is set as part of the installation process, at least on Ubuntu.

So commenting out all and then letting the `file.blockreplace` below do its job.  Works without changes for subsequent runs.

postgres/server/init.sls
```diff
@@ -184,7 +204,7 @@ postgresql-running:
   service.running:
     - name: {{ postgres.service }}
     - enable: True
-   {% if grains.os not in ('MacOS',) %}
+   {% if grains.os not in ('MacOS',) and not db_port %}
```
myii: This is the simplest way to reuse the same state but I'm sure it could be improved.  What is the right way to only restart the service when the `port` has been changed?

postgres/server/init.sls
```diff
@@ -86,23 +86,43 @@ postgresql-config-dir:
...
     {#- Detect empty values (none, '') in the config_backup #}
     - backup: {{ postgres.config_backup|default(false, true) }}
     - require:
       - file: postgresql-config-dir
+      {%- if db_port %}
+      - file: postgresql-conf-comment-port
+      {%- endif %}
     - watch_in:
-       - service: postgresql-running
+      - service: postgresql-running
```
vutny: :scream_cat:

postgres/server/init.sls
vutny: Yeah, it appears that `file.comment` state is more intelligent than I thought, and it does not comment out other strings matching same pattern when encounter the first one already commented. Interesting, I was not aware of that...

postgres/server/init.sls
```diff
@@ -184,7 +204,7 @@ postgresql-running:
   service.running:
     - name: {{ postgres.service }}
     - enable: True
-   {% if grains.os not in ('MacOS',) %}
+   {% if grains.os not in ('MacOS',) and not db_port %}
     - reload: True
```
vutny: I think we could do something like:
```yaml
     - reload: {{ not postgres.postgresconf and not db_port and not db_port == '5432' }}
```

postgres/server/init.sls
```diff
@@ -191,4 +211,12 @@ postgresql-running:
...
       - file: postgresql-pg_ident
 
+# Restart the service where reloading is not sufficient
+# Currently when changes are made to `postgresql.conf`
+{%- if postgres.postgresconf or db_port %}
+postgresql-service-restart:
+  service.running:
+    - name: {{ postgres.service }}
+{%- endif %}
+
```
vutny: @myii I like the idea of using `onchanges` requisite much better. We could do something like this.

```yaml
{%- if postgres.postgresconf or db_port %}

postgresql-service-restart:
  module.run:
    - name: service.restart
    - m_name: {{ postgres.service }}
    - onchanges:
      - file: postgresql-conf

{%- endif %}
```

myii: @vutny Of course that can be done.  I have a couple of questions:

1. I'm curious: while I've used `module.run` for my own formulas, I've been under the impression that it should only be used as a last resort, when there isn't an available state -- for my understanding, would you mind explaining the advantage of using this instead?
1. To save having to keep the Jinja (`{%- if postgres.postgresconf or db_port %}` in sync with the one being used above, how about moving this into that block above, i.e.:

```sls
{%- if postgres.postgresconf or db_port %}

postgresql-conf:
  file.blockreplace:
    - name: {{ postgres.conf_dir }}/postgresql.conf
    - marker_start: "# Managed by SaltStack: listen_addresses: please do not edit"
    - marker_end: "# Managed by SaltStack: end of salt managed zone --"
    - content: |
        {%- if postgres.postgresconf %}
        {{ postgres.postgresconf|indent(8) }}
        {%- endif %}
        {%- if db_port %}
        port = {{ db_port }}
        {%- endif %}
    - show_changes: True
    - append_if_not_found: True
    {#- Detect empty values (none, '') in the config_backup #}
    - backup: {{ postgres.config_backup|default(false, true) }}
    - require:
      - file: postgresql-config-dir
      {%- if db_port %}
      - file: postgresql-conf-comment-port
      {%- endif %}

  module.run:
    - name: service.restart
    - m_name: {{ postgres.service }}
    - onchanges:
      - file: postgresql-conf

{%- endif %}
```

Further advantages of less lines and keeping logically related states together.

aboe76: @myii and @vutny what about the apache-formula's way

instead module.run:
```
apache-reload:
  module.wait:
    - name: service.reload
    - m_name: {{ apache.service }}
```

and then have the following in the file state:
```
    - watch_in:
      - module: apache-reload
```

aboe76: or just use listen_in:
https://docs.saltstack.com/en/latest/ref/states/requisites.html#listen-listen-in

myii: @aboe76 Unfortunately, can't use `listen`/`listen_in` because the service needs to be restarted _before_ the `postgres.manage` states run.

myii: @aboe76 @vutny `module.wait` + `watch`/`watch_in` looks like an improvement, thanks for that @aboe76.  However, the other issues raised above still remain.  At the very least, where should this be positioned?  In the same `{%- if postgres.postgresconf or db_port %}` block as shown above or at the bottom next to `service.running`?

vutny: @myii I think it is OK to group together `postgresql-conf` and future `postgres-restart` states, just better name them separately, because that's a good practice.

In case of initial installation, when PG cluster is not running yet, the restart state would simply bring it up. Later `postgresql-running` state will just do reload when necessary, that's fast and safe. And it would bring the service back in case of any earlier failures. But this will prevent from unnecessary restarts then there would be no changes to `postgresql.conf` file.

With that said, your approach is good. Please update the PR and I'm sure we'll merge it in.
Thank you for your contribution!

myii:
@vutny @aboe76 So I've applied those changes using `module.wait` + `watch_in`.  Thanks for all of your feedback.

---

Using the work in this PR, I've been able to adapt the formula to work with multiple clusters/versions.  While this has far-reaching impact the implementation is relatively simple, primarily wrapping the existing code in `for` loops and appending the `port` to state/requisite identifiers.  Designed to have no effect on existing configurations, strictly opt-in.

At this stage, the implementation is rudimentary, i.e. "works for me".  Is there any appetite for including this functionality?  If there is, I'll start a new PR where it can be discussed further.

aboe76:
@vutny would you like to take a second look?

PR #222, opened 2018-06-13
Enable replication connections by default in `pg_hba.conf`

myii:
Upstream commit:

- postgres/postgres@be37c21
- 9 Mar 2017
- master REL_11_BETA1 REL_10_4 ... REL_10_BETA1

When installing version `10`, default `pg_hba.conf` additionally contains:

```conf
# Allow replication connections from localhost, by a user with the
# replication privilege.
local   replication     all                                     peer
host    replication     all             127.0.0.1/32            md5
host    replication     all             ::1/128                 md5
```

### DIFF
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -47,6 +47,11 @@ postgres:
     - ['host', 'all', 'all', '127.0.0.1/32', 'md5']
     # IPv6 local connections:
     - ['host', 'all', 'all', '::1/128', 'md5']
+    # Allow replication connections from localhost, by a user with the
+    # replication privilege.
+    - ['local', 'replication', 'all', 'peer']
+    - ['host', 'replication', 'all', '127.0.0.1/32', 'md5']
+    - ['host', 'replication', 'all', '::1/128', 'md5']
 
   pg_ident.conf: salt://postgres/templates/pg_ident.conf.j2
   identity_map: []
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -2,7 +2,7 @@
 
 postgres:
   use_upstream_repo: True
-  version: '9.5'
+  version: '10'
   pkg: postgresql
   pkgs_extra: []
   pkg_client: postgresql-client
--- a/pillar.example
+++ b/pillar.example
@@ -3,7 +3,7 @@ postgres:
   # Set True to configure upstream postgresql.org repository for YUM/APT/ZYPP
   use_upstream_repo: False
   # Version to install from upstream repository (if upstream_repo: True)
-  version: '9.6'
+  version: '10'
   # If automatic package installation fails, use `fromrepo` to specify the
   # upstream repo to install packages from [#133, #185] (if upstream_repo: True)
   fromrepo: 'jessie-pgdg'


myii:
@noelmcloughlin I've just added another commit, to separate the v10+ `acls`.  Now these are only applied when the PostgreSQL server version is 10+.  No spamming earlier versions.

Test output for version `'9.6'` (formatted):
```yaml
[
    [u'local', u'all', u'all', u'peer'],
    [u'host', u'all', u'all', u'127.0.0.1/32', u'md5'],
    [u'host', u'all', u'all', u'::1/128', u'md5']
]
```

Test output for version `'10'` (formatted):
```yaml
[
    [u'local', u'all', u'all', u'peer'],
    [u'host', u'all', u'all', u'127.0.0.1/32', u'md5'],
    [u'host', u'all', u'all', u'::1/128', u'md5'],
    [u'local', u'replication', u'all', u'peer'],
    [u'host', u'replication', u'all', u'127.0.0.1/32', u'md5'],
    [u'host', u'replication', u'all', u'::1/128', u'md5']
]
```

postgres/defaults.yaml
```diff
@@ -2,7 +2,7 @@
 
 postgres:
   use_upstream_repo: True
-  version: '9.5'
+  version: '10'
```
noelmcloughlin: Setting`10` as "default" should be okay. Debian stretch ships with version `9.6` but the `version` parameter is paired with `use_upstream_repo: True` and version `10` seems to be widely available upstream, and for [Debian stretch too](https://download.postgresql.org/pub/repos/apt/dists/stretch-pgdg/).  Approved.

postgres/defaults.yaml
```diff
@@ -47,6 +47,12 @@ postgres:
     - ['host', 'all', 'all', '127.0.0.1/32', 'md5']
     # IPv6 local connections:
     - ['host', 'all', 'all', '::1/128', 'md5']
+  acls_v10:
```
noelmcloughlin: Syntax and content looks fine. Consider changing identifier name from `acls_v10` to `acls_replication`.

postgres/server/init.sls
```diff
@@ -107,6 +107,10 @@ postgresql-conf:
 {%- endif %}
 
 {%- set pg_hba_path = salt['file.join'](postgres.conf_dir, 'pg_hba.conf') %}
+{%- set acls = postgres.acls %}
+{%- if postgres.version|to_num >= 10 %}
```
noelmcloughlin: Please use `int` instead of `to_num` (salt v2018), or even `str_to_num` (salt v2017). In this case the Jinja [builtin filters](http://jinja.pocoo.org/docs/2.10/templates/#builtin-filters) are more portable. Not everybody uses latest Salt versions in production.

postgres/defaults.yaml
vutny: I'm not sure that having multiple settings for ACL is really that good idea.
I have checked PostgreSQL docs, and it seams the replication settings are supported starting at lease from PG [release 9.0](https://www.postgresql.org/docs/9.0/static/auth-pg-hba-conf.html).

So I think it would be no harm to include these ACL rules by default for any PG version.

noelmcloughlin:
Good catch @vutny.  The purpose of the [upstream commit ](https://github.com/postgres/postgres/commit/be37c21) was to "Enable replication connections by default in pg_hba.conf" so feature already existed.  I too would prefer one `[acls]` block.

myii:
Thanks for all of the feedback. When I get a chance, I'll rebase and let you know.
------------------
Sent from my BlackBerry
________________________________

myii:
@noelmcloughlin @vutny Rebased to remove the superfluous commit.  Ready for final checks.

myii:
You're welcome.

PR #223, opened 2018-06-14
Fixed problems with per database extensions and schemas #140

t0fik:
Fixes issue #140

### DIFF
--- a/postgres/macros.jinja
+++ b/postgres/macros.jinja
@@ -12,7 +12,9 @@
 
 {%- macro format_state(name, state, kwarg) %}
 
-  {%- do kwarg.update({'name': name}) %}
+  {%- if 'name' not in kwarg %}
+    {%- do kwarg.update({'name': name}) %}
+  {%- endif %}
   {%- if 'ensure' in kwarg %}
     {%- set ensure = kwarg.pop('ensure') %}
   {%- endif %}
--- a/postgres/manage.sls
+++ b/postgres/manage.sls
@@ -44,7 +44,25 @@ postgres-reload-modules:
 # Database states
 
 {%- for name, db in postgres.databases|dictsort() %}
-
+  {%- if 'extensions' in db %}
+    {%- for ext_name, extension in db.pop('extensions')|dictsort() %}
+      {%- do extension.update({'maintenance_db': name }) %}
+{{ format_state( name + '-' + ext_name, 'postgres_extension', extension) }}
+    - require:
+      - postgres_database: postgres_database-{{ name }}
+      {%- if 'schema' in extension %}
+      - postgres_schema: postgres_schema-{{ extension.schema }}
+      {%- endif %}
+    {%- endfor %}
+  {%- endif %}
+  {%- if 'schemas' in db %}
+    {%- for schema_name, schema in db.pop('schemas')|dictsort() %}
+      {%- do schema.update({'dbname': name }) %}
+{{ format_state( schema_name, 'postgres_schema', schema) }}
+    - require:
+      - postgres_database: postgres_database-{{ name }}
+    {%- endfor %}
+  {%- endif %}
 {{ format_state(name, 'postgres_database', db) }}
   {%- if 'owner' in db or 'tablespace' in db %}
     - require:
@@ -55,7 +73,6 @@ postgres-reload-modules:
   {%- if 'tablespace' in db %}
       - postgres_tablespace: postgres_tablespace-{{ db.tablespace }}
   {%- endif %}
-
 {%- endfor %}
 
 # Schema states
@@ -63,8 +80,9 @@ postgres-reload-modules:
 {%- for name, schema in postgres.schemas|dictsort() %}
 
 {{ format_state(name, 'postgres_schema', schema) }}
-  {%- if 'owner' in schema %}
     - require:
+      - postgres_database-{{ dbname }}
+  {%- if 'owner' in schema %}
       - postgres_user: postgres_user-{{ schema.owner }}
   {%- endif %}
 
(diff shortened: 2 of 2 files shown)


postgres/manage.sls
```diff
@@ -44,7 +44,25 @@ postgres-reload-modules:
 # Database states
 
 {%- for name, db in postgres.databases|dictsort() %}
-
+  {%- if 'extensions' in db %}
+    {%- for ext_name, extension in db.pop('extensions')|dictsort() %}
+      {%- do extension.update({'maintenance_db': name }) %}
```
vutny: If we'd look at [CREATE EXTENSION](https://www.postgresql.org/docs/current/static/sql-createextension.html) PG SQL syntax, we see that there is no direct relation between an extension and a database.
So an extension is intended to be installed into a schema, and in turn a schema should be associated with a database.

The `maintenance_db` parameter for all `postges_*.present` state function means PG cluser-wide database containing PostgreSQL internal entities, this is not the database you could install extension to.
So, no need to for the line above (49).

postgres/manage.sls
```diff
@@ -44,7 +44,25 @@ postgres-reload-modules:
...
+{{ format_state( name + '-' + ext_name, 'postgres_extension', extension) }}
```
vutny: Would you mind to separate `format_state` macro call and related parameters from other Jinja flow control operators?
This would much improve the code readability.

postgres/manage.sls
```diff
@@ -44,7 +44,25 @@ postgres-reload-modules:
...
+    {%- endfor %}
+  {%- endif %}
+  {%- if 'schemas' in db %}
+    {%- for schema_name, schema in db.pop('schemas')|dictsort() %}
+      {%- do schema.update({'dbname': name }) %}
+{{ format_state( schema_name, 'postgres_schema', schema) }}
+    - require:
+      - postgres_database: postgres_database-{{ name }}
+    {%- endfor %}
+  {%- endif %}
```
vutny: Following the logic of raw SQL command, we should first create a database, then schema and finally install an existing extension. You could pop up per-DB `schemas` and `extensions` parameters from Pillar dictionary for a single database ahead, and then construct the states afterwards in a logical order.

postgres/manage.sls
```diff
@@ -55,16 +73,16 @@ postgres-reload-modules:
...
 {%- endfor %}
 
 # Schema states
 
 {%- for name, schema in postgres.schemas|dictsort() %}
 
 {{ format_state(name, 'postgres_schema', schema) }}
-  {%- if 'owner' in schema %}
     - require:
+      - postgres_database-{{ dbname }}
```
vutny: This addition would not work properly, because there are no `dbname` Jinja variable was ever assigned.
You wanted to put `schema.dbname`, right?

postgres/manage.sls
t0fik: Following salt docs for postgresql_extension.present:
maintenance_db
    Database to act on
I've checked and state behave as docs describe. Schema is optional for 'CREATE EXTENSION', default is to create extension for current database.

postgres/manage.sls
t0fik: Logic is OK. States are executed in correct order. Arrangement in file is irrelevant.

vutny: Yep. Requisite parameters simply put them in right order. That's OK.

postgres/manage.sls
vutny: Oh, yes. You're absolutely right. An extension state is little bit different from the others in this case. I have checked the code of `postgres.psql_query` function (which is the workhorse of all those states) and it simply substitutes `maintenance_db` to `psql --dbname {{maintenance_db}}` raw command.
That's quite confusing, because other functions use `dbname` parameter instead of `maintenance_db`, so I misunderstood them. Thanks for pointing me to it.

vutny: Oh, yes. You're absolutely right. An extension state is little bit different from the others in this case. I have checked the code of `postgres.psql_query` function (which is the workhorse of all those states) and it simply substitutes `maintenance_db` to `psql --dbname {{maintenance_db}}` raw command.
That's quite confusing, because other functions use `dbname` parameter instead of `maintenance_db`, so I misunderstood them. Thanks for pointing me to it.

postgres/manage.sls
vutny: Maybe you consider to put newlines between state generation code and Jinja loops?

aboe76:
@vutny can you take a second look...

postgres/macros.jinja
```diff
@@ -12,7 +12,9 @@
 
 {%- macro format_state(name, state, kwarg) %}
 
-  {%- do kwarg.update({'name': name}) %}
+  {%- if 'name' not in kwarg %}
+    {%- do kwarg.update({'name': name}) %}
+  {%- endif %}
```
vutny: This is just awesome. Would mind to update `pillar.example` file for the formula users to see they now able to provide unique IDs for PG entities?

t0fik: You can't set state id via pillar. So no need for `pillar.example` update. This feature is for formula developers only.
IMHO Allowing user setting state ids is asking for trouble.

vutny: @t0fik No-no, it is not about setting state IDs. There was PR #155 that intended to add capability to override PG entity attributes (including names) depending on some external data. But it wasn't merged. With your changes, it is also achievable.

vutny:
@aboe76 This PR is ready to merge, thanks.

aboe76:
@t0fik and @vutny thanks for this

PR #228, opened 2018-06-16
Allow conf/data directory separation

noelmcloughlin:
**This PR adds PostgreSQL support for different `data_dir` and `conf_dir` directories.**    

Supersedes  #187 which listed following advantages-

1. User may want separation of data and configuration.

2. Ubuntu postgresql-server stores config in `/etc`, and data into `/var`.  The `postgres` state fails horribly if `/etc/postgresql/../postgresql.conf` exists, perhaps after `postgres.dropped` runs. 

3. This is **current situation** where `conf_dir` variable exists but not `data_dir`.
```
codenamemap.yaml:  conf_dir: /etc/postgresql/{{ version }}/main
defaults.yaml:  conf_dir: /var/lib/pgsql/data
osmap.yaml:  conf_dir: /var/lib/postgres/data
osmap.yaml:  conf_dir: /var/lib/pgsql/{{ repo.version }}/data
osmap.yaml:  conf_dir: /usr/local/var/postgres
```

4. `postgres` install always fails after 'dropped' (or remove #182) states, because conf_dir remains.

5. If `conf_dir` and `data_dir` have same value, `pg_ctl` throws permissions error.
       `pg_ctl[8774]: FATAL:  data directory "/var/lib/pgsql/data" has group or world access`

Backwards Compatibility: Depreciated `postgres.create_cluster.command` pillar is supported.

### DIFF
--- a/postgres/codenamemap.yaml
+++ b/postgres/codenamemap.yaml
@@ -10,25 +10,26 @@
 
 {% macro debian_codename(name, version, codename=none) %}
 
-  {# use upstream version if configured #}
   {% if repo.use_upstream_repo == true %}
     {% set version = repo.version %}
     {% set fromrepo = repo.fromrepo|default(name ~ '-pgdg', true) %}
   {% else %}
     {% set fromrepo = name %}
   {% endif %}
+  {% set conf_dir = '/etc/postgresql/' ~ version ~ '/main' %}
+  {% set data_dir = '/var/lib/postgresql/' ~ version ~ '/main' %}
 
 {{ codename|default(name, true) }}:
   # PostgreSQL packages are mostly downloaded from `main` repo component
+  conf_dir: {{ conf_dir }}
+  data_dir: {{ data_dir }}
   fromrepo: {{ fromrepo }}
   pkg_repo:
     name: 'deb http://apt.postgresql.org/pub/repos/apt/ {{ name }}-pgdg main'
   pkg: postgresql-{{ version }}
   pkg_client: postgresql-client-{{ version }}
-  conf_dir: /etc/postgresql/{{ version }}/main
   prepare_cluster:
-    command: pg_createcluster {{ version }} main
-    test: test -f /var/lib/postgresql/{{ version }}/main/PG_VERSION && test -f /etc/postgresql/{{ version }}/main/postgresql.conf
+    pgcommand: pg_createcluster {{ version }} main -d
     user: root
 
 {% endmacro %}
@@ -38,15 +39,11 @@
 
   {# use upstream version if configured #}
   {% if repo.use_upstream_repo == true %}
-    {% set fromrepo = repo.fromrepo|default(name ~ '-pgdg', true) %}
     {% set version = repo.version %}
-  {% else %}
-    {% set fromrepo = name %}
   {% endif %}
 
 {{ codename|default(name, true) }}:
   # PostgreSQL packages are mostly downloaded from `main` repo component
-  fromrepo: {{ fromrepo }}
   pkg_repo:
     baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ version }}/fedora/fedora-$releasever-$basearch'
 
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -16,12 +16,13 @@ postgres:
   group: postgres
 
   prepare_cluster:
-    command: initdb --pgdata=/var/lib/pgsql/data
-    test: test -f /var/lib/pgsql/data/PG_VERSION
+    pgcommand: initdb -D
+    pgtestfile: PG_VERSION
     user: postgres
     env: []
 
   conf_dir: /var/lib/pgsql/data
+  data_dir: /var/lib/pgsql/data
   conf_dir_mode: '0700'
   postgresconf: ""
 
--- a/postgres/osfamilymap.yaml
+++ b/postgres/osfamilymap.yaml
@@ -4,9 +4,7 @@
 
 Arch:
   conf_dir: /var/lib/postgres/data
-  prepare_cluster:
-    command: initdb -D /var/lib/postgres/data
-    test: test -f /var/lib/postgres/data/PG_VERSION
+  data_dir: /var/lib/postgres/data
   pkg_client: postgresql-libs
   pkg_dev: postgresql
 
@@ -26,6 +24,8 @@ FreeBSD:
   user: pgsql
 
 OpenBSD:
+  conf_dir: /var/postgresql/data
+  data_dir: /var/postgresql/data
   user: _postgresql
 
 RedHat:
@@ -44,12 +44,9 @@ RedHat:
   pkg_client: postgresql{{ release }}
   pkg_libs: postgresql{{ release }}-libs
   pkg_dev: postgresql{{ release }}-devel
-  conf_dir: /var/lib/pgsql/{{ repo.version }}/data
+  conf_dir: {{ data_dir }}
+  data_dir: {{ data_dir }}
   service: postgresql-{{ repo.version }}
-
-  prepare_cluster:
-    command: initdb --pgdata='{{ data_dir }}'
-    test: test -f '{{ data_dir }}/PG_VERSION'
 
   # Alternatives system
   linux:
@@ -118,19 +115,16 @@ Suse:
     gpgautoimport: True
 
 {% if repo.use_upstream_repo == true %}
-  {% set lib_dir = '/var/lib/pgsql/' ~ repo.version ~ '/data' %}
+  {% set data_dir = '/var/lib/pgsql/' ~ repo.version ~ '/data' %}
 
   fromrepo: pgdg-sles-{{ release }}
   pkg: postgresql{{ release }}-server
   pkg_client: postgresql{{ release }}
   pkg_dev: postgresql{{ release }}-devel
   pkg_libs: postgresql{{ release }}-libs
-  conf_dir: {{ lib_dir }}
+  conf_dir: {{ data_dir }}
+  data_dir: {{ data_dir }}
   service: postgresql-{{ repo.version }}
-
-  prepare_cluster:
-    command: /usr/pgsql-{{ repo.version }}/bin/initdb --pgdata='{{ lib_dir }}'
-    test: test -f '{{ lib_dir }}/PG_VERSION'
 
   # Alternatives system
   linux:
@@ -184,9 +178,8 @@ MacOS:
   user: {{ repo.user }}
   group: {{ repo.group }}
   conf_dir: /Users/{{ repo.user }}/Library/AppSupport/postgres_{{ repo.use_upstream_repo }}
+  data_dir: /Users/{{ repo.user }}/Library/AppSupport/postgres_{{ repo.use_upstream_repo }}
   prepare_cluster:
-    command: initdb -D /Users/{{ repo.user }}/Library/AppSupport/postgres_{{ repo.use_upstream_repo }}
-    test: test -f /Users/{{ repo.user }}/Library/AppSupport/postgres_{{ repo.use_upstream_repo }}/PG_VERSION
     user: {{ repo.user }}
     group: {{ repo.group }}
 {%- endif %}
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -61,27 +61,43 @@ postgresql-{{ bin }}-altinstall:
 {%- endif %}
 
 postgresql-cluster-prepared:
+  file.directory:
+    - name: {{ postgres.conf_dir }}
+    - user: {{ postgres.user }}
+    - group: {{ postgres.group }}
+    - makedirs: True
+    - recurse:
+      - user
+      - group
+    - dir_mode: 755
   cmd.run:
+ {%- if postgres.prepare_cluster.command is defined %}
+      {# support for depreciated 'prepare_cluster.command' pillar #}
     - name: {{ postgres.prepare_cluster.command }}
+    - unless: {{ postgres.prepare_cluster.test }}
+ {%- else %}
+    - name: {{ postgres.prepare_cluster.pgcommand }} {{ postgres.data_dir }}
+    - unless: test -f {{ postgres.data_dir }}/{{ postgres.prepare_cluster.pgtestfile }}
+ {%- endif %}
     - cwd: /
-    - runas: {{ postgres.prepare_cluster.user }}
     - env: {{ postgres.prepare_cluster.env }}
-    - unless:
-      - {{ postgres.prepare_cluster.test }}
+    - runas: {{ postgres.user }}
     - require:
       - pkg: postgresql-server
+      - file: postgresql-cluster-prepared
 
 postgresql-config-dir:
   file.directory:
-    - name: {{ postgres.conf_dir }}
+    - names:
+      - {{ postgres.data_dir }}
+      - {{ postgres.conf_dir }}
     - user: {{ postgres.user }}
     - group: {{ postgres.group }}
     - dir_mode: {{ postgres.conf_dir_mode }}
     - force: True
     - file_mode: 644
     - recurse:
-      - user
-      - group
+      - mode
     - makedirs: True
     - require:
       - cmd: postgresql-cluster-prepared
--- a/postgres/dropped.sls
+++ b/postgres/dropped.sls
@@ -27,4 +27,6 @@ postgresql-removed:
 
 postgres-dir-absent:
   file.absent:
-    - name: {{ postgres.conf_dir }}
+    - names:
+      - {{ postgres.conf_dir }}
+      - {{ postgres.data_dir }}
--- a/postgres/server/image.sls
+++ b/postgres/server/image.sls
@@ -12,10 +12,10 @@ include:
 
 postgresql-start:
   cmd.run:
-    - name: pg_ctl -D {{ postgres.conf_dir }} -l logfile start
+    - name: pg_ctl -D {{ postgres.data_dir }} -l logfile start
     - runas: {{ postgres.user }}
     - unless:
-      - ps -p $(head -n 1 {{ postgres.conf_dir }}/postmaster.pid) 2>/dev/null
+      - ps -p $(head -n 1 {{ postgres.data_dir }}/postmaster.pid) 2>/dev/null
     - require:
       - file: postgresql-pg_hba
 


noelmcloughlin:
Sample output on Ubuntu
--
```
         ID: postgresql-repo
    Function: pkgrepo.managed
        Name: deb http://apt.postgresql.org/pub/repos/apt xenial-pgdg main 9.5
      Result: True
     Comment: Package repo 'deb http://apt.postgresql.org/pub/repos/apt xenial-pgdg main 9.5' already configured
     Started: 06:00:11.405949
    Duration: 111.208 ms
     Changes:   
----------
          ID: postgresql-server
    Function: pkg.installed
      Result: True
     Comment: All specified packages are already installed
     Started: 06:00:12.249551
    Duration: 800.838 ms
     Changes:   
----------
    Function: file.directory
        Name: /etc/postgresql/9.5/main
      Result: True
     Comment: Directory /etc/postgresql/9.5/main updated
     Started: 06:00:13.054608
    Duration: 3.415 ms
     Changes:   
              ----------
              /etc/postgresql/9.5/main:
                  New Dir
----------
    Function: cmd.run
        Name: pg_createcluster 9.5 main -d  /var/lib/postgresql/9.5/main
      Result: True
     Comment: Command "pg_createcluster 9.5 main -d  /var/lib/postgresql/9.5/main" run
     Started: 06:00:13.059996
    Duration: 11420.288 ms
     Changes:   
              ----------
              pid:
                  597064
              retcode:
                  0
              stderr:
              stdout:
                  Creating new PostgreSQL cluster 9.5/main ...
                  /usr/lib/postgresql/9.5/bin/initdb -D /var/lib/postgresql/9.5/main --auth-local peer --auth-host md5
                  The files belonging to this database system will be owned by user "postgres".
                  This user must also own the server process.
                  
                  The database cluster will be initialized with locale "C".
                  The default database encoding has accordingly been set to "SQL_ASCII".
                  The default text search configuration will be set to "english".
                  
                  Data page checksums are disabled.
                  
                  fixing permissions on existing directory /var/lib/postgresql/9.5/main ... ok
                  creating subdirectories ... ok
                  selecting default max_connections ... 100
                  selecting default shared_buffers ... 128MB
                  selecting dynamic shared memory implementation ... posix
                  creating configuration files ... ok
                  creating template1 database in /var/lib/postgresql/9.5/main/base/1 ... ok
                  initializing pg_authid ... ok
                  initializing dependencies ... ok
                  creating system views ... ok
                  loading system objects' descriptions ... ok
                  creating collations ... ok
                  creating conversions ... ok
                  creating dictionaries ... ok
                  setting privileges on built-in objects ... ok
                  creating information schema ... ok
                  loading PL/pgSQL server-side language ... ok
                  vacuuming database template1 ... ok
                  copying template1 to template0 ... ok
                  copying template1 to postgres ... ok
                  syncing data to disk ... ok
                  
                  Success. You can now start the database server using:
                  
                      /usr/lib/postgresql/9.5/bin/pg_ctl -D /var/lib/postgresql/9.5/main -l logfile start
                  
                  Ver Cluster Port Status Owner    Data directory               Log file
                  9.5 main    5432 down   postgres /var/lib/postgresql/9.5/main /var/log/postgresql/postgresql-9.5-main.log
----------
          ID: postgresql-config-dir
    Function: file.directory
        Name: /var/lib/postgresql/9.5/main
      Result: True
     Comment: Directory /var/lib/postgresql/9.5/main updated
     Started: 06:00:24.481635
    Duration: 971.997 ms
     Changes:   
              ----------
              mode:
                  0644
----------
          ID: postgresql-config-dir
    Function: file.directory
        Name: /etc/postgresql/9.5/main
      Result: True
     Comment: Directory /etc/postgresql/9.5/main updated
     Started: 06:00:25.454164
    Duration: 5.309 ms
     Changes:   
              ----------
              mode:
                  0644
----------
          ID: postgresql-conf
    Function: file.blockreplace
        Name: /etc/postgresql/9.5/main/postgresql.conf
      Result: True
     Comment: Changes were made
     Started: 06:00:25.460286
    Duration: 9.067 ms
     Changes:   
              ----------
              diff:
                  --- 
                  +++ 
                  @@ -623,3 +623,6 @@
                   #------------------------------------------------------------------------------
                   
                   # Add settings for extensions here
                  +# Managed by SaltStack: listen_addresses: please do not edit
                  +listen_addresses = '*'  # listen on all interfaces
                  +# Managed by SaltStack: end of salt managed zone --
----------
          ID: postgresql-pg_hba
    Function: file.managed
        Name: /etc/postgresql/9.5/main/pg_hba.conf
      Result: True
     Comment: File /etc/postgresql/9.5/main/pg_hba.conf updated
     Started: 06:00:25.470057
    Duration: 43826.111 ms
     Changes:   
              ----------
              diff:
                  --- 
                  +++ 
                  @@ -1,99 +1,31 @@
                  +######################################################################
                  +# ATTENTION! Managed by SaltStack.                                   #
                  +# DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN! #
                  +######################################################################
                  +#
                   # PostgreSQL Client Authentication Configuration File
                   # ===================================================
                   #
                   # Refer to the "Client Authentication" section in the PostgreSQL
                  -# documentation for a complete description of this file.  A short
                  -# synopsis follows.
                  -#
                  -# This file controls: which hosts are allowed to connect, how clients
                  -# are authenticated, which PostgreSQL user names they can use, which
                  -# databases they can access.  Records take one of these forms:
                  -#
                  -# local      DATABASE  USER  METHOD  [OPTIONS]
                  -# host       DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
                  -# hostssl    DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
                  -# hostnossl  DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
                  -#
                  -# (The uppercase items must be replaced by actual values.)
                  -#
                  -# The first field is the connection type: "local" is a Unix-domain
                  -# socket, "host" is either a plain or SSL-encrypted TCP/IP socket,
                  -# "hostssl" is an SSL-encrypted TCP/IP socket, and "hostnossl" is a
                  -# plain TCP/IP socket.
                  -#
                  -# DATABASE can be "all", "sameuser", "samerole", "replication", a
                  -# database name, or a comma-separated list thereof. The "all"
                  -# keyword does not match "replication". Access to replication
                  -# must be enabled in a separate record (see example below).
                  -#
                  -# USER can be "all", a user name, a group name prefixed with "+", or a
                  -# comma-separated list thereof.  In both the DATABASE and USER fields
                  -# you can also write a file name prefixed with "@" to include names
                  -# from a separate file.
                  -#
                  -# ADDRESS specifies the set of hosts the record matches.  It can be a
                  -# host name, or it is made up of an IP address and a CIDR mask that is
                  -# an integer (between 0 and 32 (IPv4) or 128 (IPv6) inclusive) that
                  -# specifies the number of significant bits in the mask.  A host name
                  -# that starts with a dot (.) matches a suffix of the actual host name.
                  -# Alternatively, you can write an IP address and netmask in separate
                  -# columns to specify the set of hosts.  Instead of a CIDR-address, you
                  -# can write "samehost" to match any of the server's own IP addresses,
                  -# or "samenet" to match any address in any subnet that the server is
                  -# directly connected to.
                  -#
                  -# METHOD can be "trust", "reject", "md5", "password", "gss", "sspi",
                  -# "ident", "peer", "pam", "ldap", "radius" or "cert".  Note that
                  -# "password" sends passwords in clear text; "md5" is preferred since
                  -# it sends encrypted passwords.
                  -#
                  -# OPTIONS are a set of options for the authentication in the format
                  -# NAME=VALUE.  The available options depend on the different
                  -# authentication methods -- refer to the "Client Authentication"
                  -# section in the documentation for a list of which options are
                  -# available for which authentication methods.
                  -#
                  -# Database and user names containing spaces, commas, quotes and other
                  -# special characters must be quoted.  Quoting one of the keywords
                  -# "all", "sameuser", "samerole" or "replication" makes the name lose
                  -# its special character, and just match a database or username with
                  -# that name.
                  -#
                  -# This file is read on server startup and when the postmaster receives
                  -# a SIGHUP signal.  If you edit the file on a running system, you have
                  -# to SIGHUP the postmaster for the changes to take effect.  You can
                  -# use "pg_ctl reload" to do that.
                  -
                  -# Put your actual configuration here
                  -# ----------------------------------
                  -#
                  -# If you want to allow non-local connections, you need to add more
                  -# "host" records.  In that case you will also need to make PostgreSQL
                  -# listen on a non-local interface via the listen_addresses
                  -# configuration parameter, or via the -i or -h command line switches.
                  -
                  -
                  -
                  +# documentation for a complete description of this file.
                   
                   # DO NOT DISABLE!
                   # If you change this first entry you will need to make sure that the
                   # database superuser can access the database using some other method.
                   # Noninteractive access to all databases is required during automatic
                   # maintenance (custom daily cronjobs, replication, and similar tasks).
                  -#
                  +
                   # Database administrative login by Unix domain socket
                   local   all             postgres                                peer
                   
                   # TYPE  DATABASE        USER            ADDRESS                 METHOD
                  -
                  -# "local" is for Unix domain socket connections only
                  -local   all             all                                     peer
                  -# IPv4 local connections:
                  -host    all             all             127.0.0.1/32            md5
                  -# IPv6 local connections:
                  -host    all             all             ::1/128                 md5
                  -# Allow replication connections from localhost, by a user with the
                  -# replication privilege.
                  -#local   replication     postgres                                peer
                  -#host    replication     postgres        127.0.0.1/32            md5
                  -#host    replication     postgres        ::1/128                 md5
                  +local   db1             localUser                               md5
                  +host    db2             remoteUser      192.168.33.0/24         md5
                  +local   all             postgres                                ident
                  +local   citrix-confdb   root                                    ident
                  +local   citrix-confdb   ctxsrvr                                 ident
                  +local   citrix-confdb   guest                                   trust
                  +host    citrix-confdb   ctxvda          127.0.0.1/32            password
                  +host    citrix-confdb   guest           127.0.0.1/32            trust
                  +host    citrix-confdb   ctxvda          ::1/128                 password
                  +host    citrix-confdb   guest           ::1/128                 trust
              mode:
                  0600
----------
          ID: postgresql-pg_ident
    Function: file.managed
        Name: /etc/postgresql/9.5/main/pg_ident.conf
      Result: True
     Comment: 
     Started: 06:01:09.297097
    Duration: 2.776 ms
     Changes:   
              ----------
              mode:
                  0600
----------
          ID: postgresql-tablespace-dir-my_space
    Function: file.directory
        Name: /srv/my_tablespace
      Result: True
     Comment: Directory /srv/my_tablespace is in the correct state
              Directory /srv/my_tablespace updated
     Started: 06:01:09.300458
    Duration: 261.818 ms
     Changes:   
----------
          ID: postgresql-running
    Function: service.running
        Name: postgresql
      Result: True
     Comment: Service reloaded
     Started: 06:01:09.642742
    Duration: 135.383 ms
     Changes:   
              ----------
              postgresql:
                  True
```

postgres/codenamemap.yaml
```diff
@@ -14,21 +15,26 @@
   {% if repo.use_upstream_repo == true %}
     {% set version = repo.version %}
     {% set fromrepo = repo.fromrepo|default(name ~ '-pgdg', true) %}
+    {% set conf_dir = '/etc/postgresql/' + version|string + '/main' %}
+    {% set data_dir = '/var/lib/postgresql/' + version|string + '/main' %}
   {% else %}
     {% set fromrepo = name %}
+    {% set conf_dir = defaults.conf_dir %}
+    {% set data_dir = defaults.data_dir %}
```
vutny: I have checked Ubuntu docs on [PostgreSQL](https://help.ubuntu.com/lts/serverguide/postgresql.html), and it appears that "downstream" packages also store all configuration files in `/etc` as well. Tried to install PG on Debian: same thing there.

So we need to separate conf and data directory for the whole Debian OS family.

postgres/codenamemap.yaml
```diff
@@ -40,13 +46,19 @@
   {% if repo.use_upstream_repo == true %}
     {% set fromrepo = repo.fromrepo|default(name ~ '-pgdg', true) %}
     {% set version = repo.version %}
+    {% set conf_dir = '/etc/postgresql/' + version|string + '/main' %}
+    {% set data_dir = '/var/lib/postgresql/' + version|string + '/main' %}
```
vutny: Is this really applicable to Fedora? I'm not sure.

postgres/codenamemap.yaml
```diff
@@ -40,13 +46,19 @@
   {% if repo.use_upstream_repo == true %}
     {% set fromrepo = repo.fromrepo|default(name ~ '-pgdg', true) %}
```
vutny: I guess I missed this before, but the `fromrepo` for RedHat family should be equal to `pkg_repo.name` value (`pgdg{{ release }}`). This need to be corrected.

postgres/server/init.sls
```diff
@@ -61,27 +61,43 @@ postgresql-{{ bin }}-altinstall:
...
+      - user
+      - group
+    - dir_mode: 755
   cmd.run:
+ {%- if postgres.prepare_cluster.command is defined %}
+      {# support for depreciated 'prepare_cluster.command' pillar #}
     - name: {{ postgres.prepare_cluster.command }}
+    - unless: {{ postgres.prepare_cluster.test }}
+ {%- else %}
+    - name: {{ postgres.prepare_cluster.pgcommand + ' ' }} {{ postgres.data_dir }}
```
vutny: I saw you have reconstructed the `pgcommand` defaults... I like that.
But just consider to use short option for all platforms except Debians: `initdb -D`. In such case adding that space in the template would be unnecessary.

postgres/defaults.yaml
```diff
@@ -15,12 +15,13 @@ postgres:
   group: postgres
 
   prepare_cluster:
-    command: initdb --pgdata=/var/lib/pgsql/data
-    test: test -f /var/lib/pgsql/data/PG_VERSION
+    pgcommand: initdb --pgdata=
```
vutny: Having `initdb -D` here will simplify your final command for the `postgresql-cluster-prepared` state.

postgres/osfamilymap.yaml
```diff
@@ -4,9 +4,10 @@
 
 Arch:
   conf_dir: /var/lib/postgres/data
+  data_dir: /var/lib/postgres/data
   prepare_cluster:
-    command: initdb -D /var/lib/postgres/data
-    test: test -f /var/lib/postgres/data/PG_VERSION
+    pgcommand: initdb -D
+    pgtestfile: PG_VERSION
```
vutny: Since these values would exactly match the default ones, no need to override them here. Same for other distros below.

postgres/codenamemap.yaml
noelmcloughlin: I checked my commits on #187 (which was tested heavily) and you're correct. Fixed.

postgres/codenamemap.yaml
noelmcloughlin: I checked my commits on #187 (which was tested heavily) and you're correct. Not needed. Fixed.

postgres/server/init.sls
noelmcloughlin: So I can replace `{{ postgres.prepare_cluster.pgcommand + ' ' }}` with `initdb -D` for all except Debian?

postgres/osfamilymap.yaml
noelmcloughlin: Good point.  I'll fix that. thanks @vutny

postgres/defaults.yaml
noelmcloughlin: Cool!! I'm only realizing now `initdb -D` and `initdb --pgdata=` are equilivent. ;-)

postgres/codenamemap.yaml
```diff
@@ -10,25 +11,26 @@
...
   {% else %}
     {% set fromrepo = name %}
   {% endif %}
+  {% set conf_dir = '/etc/postgresql/' + version|string + '/main' %}
+  {% set data_dir = '/var/lib/postgresql/' + version|string + '/main' %}
 
 {{ codename|default(name, true) }}:
   # PostgreSQL packages are mostly downloaded from `main` repo component
+  conf_dir: {{ conf_dir }}
+  data_dir: {{ data_dir }}
```
vutny: Since these values become global for all Debian derivatives, we should move them to the OS family map.

postgres/codenamemap.yaml
```diff
@@ -46,7 +48,7 @@
 
 {{ codename|default(name, true) }}:
   # PostgreSQL packages are mostly downloaded from `main` repo component
-  fromrepo: {{ name }}
+  fromrepo: {{ fromrepo }}
```
vutny: This really needs to be fixed. The `fromrepo` for "downstream" installations should be `None` by default, but for the upstream it should match the `pkg_repo.name` value across RedHat-based distros. I see that it is already done at upper mapping level in the OS family map. So we can just remove manipulations on `fromrepo` here.

postgres/codenamemap.yaml
```diff
@@ -1,6 +1,7 @@
 ### Set parameters based on PostgreSQL version supplied with particular distro
 
 {% import_yaml "postgres/repo.yaml" as repo %}
+{% import_yaml "postgres/defaults.yaml" as defaults %}
```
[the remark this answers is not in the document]
noelmcloughlin: yes, thanks.

postgres/codenamemap.yaml
noelmcloughlin: The `version` calculated by oscodename is passed to `fedora_codename` macro and used  ..
```
  {% set conf_dir = '/etc/postgresql/' + version|string + '/main' %}
  {% set data_dir = '/var/lib/postgresql/' + version|string + '/main' %}
```
If version calculated by OS map is different to version calculated here, they maybe problems?  What do you think?

postgres/codenamemap.yaml
noelmcloughlin: Yes, its handled correctly for "upstream" repos in OS family map file.  The `fromrepo: ''` is defaults.yaml should evaluate to None,  for Debian it should be `name` but for Fedora/Redhat what should the "downstream" fromrepo value be?

vutny: I think we should we that default and does not handle the `fromrepo` parameter for downstream at all. That's rarely applicable usecase and would be to complex to handle for now.

postgres/codenamemap.yaml
vutny: Oh, you're right, absolutely. I've completely forgot that the version for distro-provided packages is handled here as well. Thanks, keep insisting on right things! :+1:

postgres/codenamemap.yaml
```diff
@@ -10,25 +11,26 @@
...
 
-  {# use upstream version if configured #}
   {% if repo.use_upstream_repo == true %}
     {% set version = repo.version %}
     {% set fromrepo = repo.fromrepo|default(name ~ '-pgdg', true) %}
   {% else %}
     {% set fromrepo = name %}
   {% endif %}
+  {% set conf_dir = '/etc/postgresql/' + version|string + '/main' %}
+  {% set data_dir = '/var/lib/postgresql/' + version|string + '/main' %}
```
vutny: That's OK, but generally better to use Jinja concatenation operator `~` (tilde) or `file.join` Salt's function here instead of type casting on variables.

postgres/codenamemap.yaml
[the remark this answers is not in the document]
vutny: Don't forget about it :smiley:

postgres/codenamemap.yaml
noelmcloughlin: Forgot about tilde - very useful operand in this case. thx

> ~
Converts all operands into strings and concatenates them.

postgres/codenamemap.yaml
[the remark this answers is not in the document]
noelmcloughlin: ha ha - think I nailed it this time!! ;-)

noelmcloughlin:
One question. If we choose non-standard `data_dir` or `conf_dir` will Apparmor block PG? See https://blogs.oracle.com/jsmyth/apparmor-and-mysql.  I cannot find any documented issue and saw no issue in testing.  Did not test in Docker however.

postgres/codenamemap.yaml
noelmcloughlin: Okay, so `fromrepo` is not used for Fedora. For Ubuntu I do need to set `fromrepo` jinja variable to some value, either `{% set fromrepo = name %}` or `{% set fromrepo = '' %}` or `{% set fromrepo = None %}` in codenamemap.yaml.  The SLS files will check `{% if postgres.fromrepo }` so just need to set correct value so condition fails.  Is this how you see it too?

aboe76:
@noelmcloughlin and @vutny merged!

noelmcloughlin:
Cool thanks!!

PR #206, opened 2018-04-09
Convert hex to decimal in defaults.yaml

noelmcloughlin:
Fixes issue.
```
[INFO    ] Executing state file.directory for [/etc/postgresql/9.5/main]
[ERROR   ] Failed to change mode to 0509```

PR #231, opened 2018-06-21
Fix `conf_dir` => `data_dir` required for preparing cluster

myii:
@noelmcloughlin @vutny Thanks for #228, that came at the right time for me as well.  However, this fix is necessary, since the `data_dir` needs to be in place before preparing the cluster.

I've only done basic testing but there also seems to be an issue with starting the server -- otherwise the `postgres.manage` states fail.  I resolved this by adding `--start` to the `pg_createcluster` command.  What are your thoughts about this?

### DIFF
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -85,6 +85,8 @@ postgresql-cluster-prepared:
     - require:
       - pkg: postgresql-server
       - file: postgresql-cluster-prepared
+    - watch_in:
+      - module: postgresql-service-restart
 
 postgresql-config-dir:
   file.directory:
@@ -140,14 +142,14 @@ postgresql-conf:
     - watch_in:
       - module: postgresql-service-restart
 
+{%- endif %}
+
 # Restart the service where reloading is not sufficient
-# Currently only when changes are made to `postgresql.conf`
+# Currently when the cluster is created or changes made to `postgresql.conf`
 postgresql-service-restart:
   module.wait:
     - name: service.restart
     - m_name: {{ postgres.service }}
-
-{%- endif %}
 
 {%- set pg_hba_path = salt['file.join'](postgres.conf_dir, 'pg_hba.conf') %}
 


myii:
OK, I've just rebased the commit because the regression in #228 is:

* `runas: {{ postgres.prepare_cluster.user }}` changed to `{{ postgres.user }}`

First error encountered, prompting this PR (Ubuntu minion):

```salt
stderr:
    install: cannot create directory '/var/lib/postgresql': Permission denied
    Error: could not create data directory; you might need to run this program with root privileges
```

Another error, even after the initial fix (fresh PostgreSQL installation):

```salt
stderr:
    Error: could not create log directory; you might need to run this program with root privileges
```

myii:
I've just added another commit in relation to the server starting before `postgres.manage`.  Simply taking advantage of the `service.restart` by moving it out of the `if` block (for `postgresql.conf` changes) and adding the `watch_in`.

postgres/server/init.sls
```diff
@@ -81,7 +81,7 @@ postgresql-cluster-prepared:
  {%- endif %}
     - cwd: /
     - env: {{ postgres.prepare_cluster.env }}
-    - runas: {{ postgres.user }}
+    - runas: {{ postgres.prepare_cluster.user }}
```
noelmcloughlin: Yes, the previous value was `postgres.prepare_cluster.user` so we should revert.

myii:
@noelmcloughlin No problem, thanks for the review.

noelmcloughlin:
One general comment regarding `prepare_cluster` states.  PR #202, #206 and #210 (& others) tried to deal with the *permission* problem - we do want `0700` permission on `data_dir` - but it remains critically important that `data_dir` has correct ownership (thanks for your fix)  because the "prepare cluster" command is **[runas](https://github.com/myii/postgres-formula/blob/67d6b8a5450906669bba6cea33253d03b228d4ec/postgres/server/init.sls#L84)** non-root.  Hopefully we are closer to full solution with this PR.

myii:
@noelmcloughlin The main progress was yours in #228 but glad to have been of service.

>  ... because the "prepare cluster" command is runas non-root ...

Except for Debian-based systems of course (due to `pg_createcluster`), where `postgres.prepare_cluster.user` evaluates to `root`:

https://github.com/saltstack-formulas/postgres-formula/blob/9865664963bb6c5a07dc306caa9e9947aecdab15/postgres/codenamemap.yaml#L31-L33

noelmcloughlin:
@myii  Good point.  Even on Debian, the user/command values [revert](https://github.com/saltstack-formulas/postgres-formula/blob/9865664963bb6c5a07dc306caa9e9947aecdab15/postgres/defaults.yaml#L18) to `postgres` and  `initdb -D` whenever `use_upstream_repo: True` is set.

myii:
@noelmcloughlin I'm not finding that at my end during a test run:

```yaml
...

prepare_cluster:
  env: []
  pgcommand: pg_createcluster 10 testing07 -d
  pgtestfile: PG_VERSION
  user: root

...

use_upstream_repo: true
```

PR #232, opened 2018-06-21
Allow configuration of cluster `name`, `locale` & `encoding`

myii:
#### Commit: Allow configuration of cluster `name`, `locale` & `encoding`

* `name` replaces literal string `main`
* `locale` and `encoding` currently default to `C` and `SQL_ASCII` respectively -- obvious benefit to being able to configure these
    * Using the long form switches for both of these (i.e. `--locale=` & `--encoding=`), which have identical usage in both `initdb` and `pg_createcluster`

---

#### Commit: Remove trailing slash so `pkgrepo.absent` works for Apt

* Trailing slash gets removed by `pkgrepo.managed` so then not found by `pkgrepo.absent`

---

#### Commit: Fix `pillar.get` for `MacOS` user/group early lookup

* Noticed this while working on this PR
* Untested but surely the periods need to be replaced by colons, right?
* If not, will rebase to remove this commit

### DIFF
--- a/postgres/repo.yaml
+++ b/postgres/repo.yaml
@@ -6,6 +6,7 @@
 use_upstream_repo: {{ salt['pillar.get']('postgres:use_upstream_repo', defaults.postgres.use_upstream_repo) }}
 version: {{ salt['pillar.get']('postgres:version', defaults.postgres.version) }}
 fromrepo: {{ salt['pillar.get']('postgres:fromrepo', defaults.postgres.fromrepo) }}
+cluster_name: {{ salt['pillar.get']('postgres:cluster:name', defaults.postgres.cluster.name) }}
 
 #Early lookup for system user on MacOS
 {% if grains.os == 'MacOS' %}
--- a/postgres/repo.yaml
+++ b/postgres/repo.yaml
@@ -10,8 +10,8 @@ cluster_name: {{ salt['pillar.get']('postgres:cluster:name', defaults.postgres.c
 
 #Early lookup for system user on MacOS
 {% if grains.os == 'MacOS' %}
-  {% set sysuser = salt['pillar.get']('postgres.user', salt['cmd.run']("stat -f '%Su' /dev/console")) %}
-  {% set sysgroup = salt['pillar.get']('postgres.group', salt['cmd.run']("stat -f '%Sg' /dev/console")) %}
+  {% set sysuser = salt['pillar.get']('postgres:user', salt['cmd.run']("stat -f '%Su' /dev/console")) %}
+  {% set sysgroup = salt['pillar.get']('postgres:group', salt['cmd.run']("stat -f '%Sg' /dev/console")) %}
 user: {{ sysuser }}
 group: {{ sysgroup }}
 {% endif %}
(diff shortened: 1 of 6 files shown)


myii:
#### Sample output

Based on pillar data:

```salt
postgres:
  version: '10'
  cluster:
    name: testing07
    locale: en_GB.UTF-8
    encoding: UTF8
```

Result:

```salt
Function: cmd.run
    Name: pg_createcluster 10 testing07 -d /var/lib/postgresql/10/testing07 --locale=en_GB.UTF-8 --encoding=UTF8
  Result: True
 Comment: Command "pg_createcluster 10 testing07 -d /var/lib/postgresql/10/testing07 --locale=en_GB.UTF-8 --encoding=UTF8" run
 Started: 07:58:59.222295
Duration: 3380.625 ms
 Changes:   
          ----------
          pid:
              22103
          retcode:
              0
          stderr:
          stdout:
              Creating new PostgreSQL cluster 10/testing07 ...
              /usr/lib/postgresql/10/bin/initdb -D /var/lib/postgresql/10/testing07 --auth-local peer --auth-host md5 --encoding UTF8 --locale en_GB.UTF-8
              The files belonging to this database system will be owned by user "postgres".
              This user must also own the server process.

              The database cluster will be initialized with locale "en_GB.UTF-8".
              The default text search configuration will be set to "english".
```

postgres/repo.yaml
```diff
@@ -6,11 +6,12 @@
 use_upstream_repo: {{ salt['pillar.get']('postgres:use_upstream_repo', defaults.postgres.use_upstream_repo) }}
 version: {{ salt['pillar.get']('postgres:version', defaults.postgres.version) }}
 fromrepo: {{ salt['pillar.get']('postgres:fromrepo', defaults.postgres.fromrepo) }}
+cluster_name: {{ salt['pillar.get']('postgres:cluster:name', defaults.postgres.cluster.name) }}
 
 #Early lookup for system user on MacOS
 {% if grains.os == 'MacOS' %}
-  {% set sysuser = salt['pillar.get']('postgres.user', salt['cmd.run']("stat -f '%Su' /dev/console")) %}
-  {% set sysgroup = salt['pillar.get']('postgres.group', salt['cmd.run']("stat -f '%Sg' /dev/console")) %}
+  {% set sysuser = salt['pillar.get']('postgres:user', salt['cmd.run']("stat -f '%Su' /dev/console")) %}
```
noelmcloughlin: nice catch.

postgres/server/init.sls
```diff
@@ -76,7 +76,14 @@ postgresql-cluster-prepared:
     - name: {{ postgres.prepare_cluster.command }}
     - unless: {{ postgres.prepare_cluster.test }}
  {%- else %}
-    - name: {{ postgres.prepare_cluster.pgcommand }} {{ postgres.data_dir }}
+    {%- set cc_cmd = '{0} {1}'.format(postgres.prepare_cluster.pgcommand, postgres.data_dir) %}
```
noelmcloughlin: Nice. Locale is unset by default. So if value is set thats means locale pillar was set.

Could you move this Jinja logic into `map.jinja` instead of `init.sls` file?

myii:
@noelmcloughlin Thanks for the review.  Added another commit to move the logic to `map.jinja`.

myii:
You're welcome, @noelmcloughlin.  Anything else required for the merge?

aboe76:
@myii thanks for this, @noelmcloughlin thanks for reviewing.

myii:
Thanks @aboe76.

Issue #241, opened 2018-10-01
jinja strip white spaces option in salt master config causes a broken pg_hba.conf


bebosudo:
Hi,
I'm using this formula, along with the following lines in my salt master configuration:
```
jinja_env:
  trim_blocks: True
  lstrip_blocks: True
```
to strip empty lines in my salt states. This causes the following `pg_hba.conf` file:
```
          ID: postgresql-pg_hba
    Function: file.managed
        Name: /var/lib/pgsql/data/pg_hba.conf
      Result: True
     Comment: File /var/lib/pgsql/data/pg_hba.conf updated
     Started: 10:18:10.901757
    Duration: 6342.325 ms
     Changes:   
              ----------
              diff:
                  --- 
                  +++ 
                  @@ -18,6 +18,4 @@
                   # Database administrative login by Unix domain socket
                   local   all             postgres                                peer
                   
                  -# TYPE  DATABASE        USER            ADDRESS                 METHOD
                  -host    all             all             127.0.0.1/32            md5
                  -host    all             all             ::1/128                 md5
                  +# TYPE  DATABASE        USER            ADDRESS                 METHODhost    all             all             127.0.0.1/32            md5host    all             all             ::1/128                 md5
```

Which is caused by lines https://github.com/saltstack-formulas/postgres-formula/blob/87f6cd880ca637984268e3bbe4ef7078f47ced08/postgres/templates/pg_hba.conf.j2#L39 and L23.
Do you think a pull request to remove the line strip in jinja could be feasible?

javierbertoli:
Fixed on #244

myii:
#244 results in a side effect for those running with the default values of `False` for the Jinja block trimming/stripping environment variables:

```diff
 # TYPE  DATABASE        USER            ADDRESS                 METHOD
+
+
 local   all             all                                     peer
+
 host    all             all             127.0.0.1/32            md5
+
 host    all             all             ::1/128                 md5
+
 local   replication     all                                     peer
+
 host    replication     all             127.0.0.1/32            md5
+
 host    replication     all             ::1/128                 md5
```

* Blank lines are inserted for all existing `pg_hba.conf` files.

Of course, the situation resolved by #244 is critical; but this side effect is an annoyance.  Is there any way (similar to `config.get`) to check for these Jinja values and apply the fix for either setting?

CC: @vutny.

PR #244, opened 2018-10-03
Edit pg_hba template to work with jinja 'trim whitespaces' option on salt master

bebosudo:
This should solve #241.


javierbertoli:
Thanks @bebosudo & @noelmcloughlin

PR #245, opened 2018-10-09
Make the preparation/initdb optional

gilou:
In some situations it's nice not to require the cluster to be prepared, e.g. when we are working with standbys that will be synchronized using repmgr. This makes it an option, that is set to True by default to make the state work the same way it used to.

This could blso be an occasion to add prepare_cluster examples in the pillar file.


aboe76:
@gilou and @noelmcloughlin thanks for this.

PR #246, opened 2018-10-10
allow to add bin_dir to path 

gilou:
When installed from repo, add a add_path option to include bin_dir in $PATH through a file in /etc/profile.d
Disabled by default, as it wasn't added before.


gilou:
I agree for the name, maybe it can be extended, and I wasn't thrilled by my choice. Is the explanation I added clear enough?

javierbertoli:
Merged. Thanks, @gilou! :+1:

PR #256, opened 2019-03-22
Use copy state instead of salt-call

thomasrossetto:
Now it works also with salt-ssh.

### DIFF
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -168,8 +168,9 @@ postgresql-pg_hba:
 {%- else %}
     - replace: False
 {%- endif %}
-    - require:
+    - prereq:
       - file: backup_pg_hba
+    - require:
       - file: postgresql-config-dir
     - watch_in:
       - service: postgresql-running
@@ -194,8 +195,9 @@ postgresql-pg_ident:
 {%- else %}
     - replace: False
 {%- endif %}
-    - require:
+    - prereq:
       - file: backup_pg_ident
+    - require:
       - file: postgresql-config-dir
       {%- if postgres.prepare_cluster.run %}
       - cmd: postgresql-cluster-prepared
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -164,14 +164,12 @@ postgresql-pg_hba:
     # Create the empty file before managing to overcome the limitation of check_cmd
     - onlyif: test -f {{ pg_hba_path }} || touch {{ pg_hba_path }}
     # Make a local backup before the file modification
-    - check_cmd: >-
-        salt-call --local file.copy
-        {{ pg_hba_path }} {{ pg_hba_path ~ postgres.config_backup }} remove_existing=true
   {%- endif %}
 {%- else %}
     - replace: False
 {%- endif %}
     - require:
+      - file: backup_pg_hba
       - file: postgresql-config-dir
     - watch_in:
       - service: postgresql-running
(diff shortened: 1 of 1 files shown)


thomasrossetto:
Hi @vutny !
I tried to replace with `prereq` instead of `require`and with salt-ssh i don't have any problems.
But, i don't able to test if this is the correct replace (according to your CR) for salt-call command.

vutny:
No worries @thomasrossetto !

Just update the PR with your new changes and I think we'd make it through.
Thank you for your efforts.

postgres/server/init.sls
```diff
@@ -164,13 +164,12 @@ postgresql-pg_hba:
     # Create the empty file before managing to overcome the limitation of check_cmd
     - onlyif: test -f {{ pg_hba_path }} || touch {{ pg_hba_path }}
```
vutny: This is no longer needed since we are going to remove `check_cmd` parameter. Same applies for the `pg_ident`.

postgres/server/init.sls
```diff
@@ -164,13 +164,12 @@ postgresql-pg_hba:
...
     # Make a local backup before the file modification
-    - check_cmd: >-
-        salt-call --local file.copy
-        {{ pg_hba_path }} {{ pg_hba_path ~ postgres.config_backup }} remove_existing=true
   {%- endif %}
 {%- else %}
     - replace: False
 {%- endif %}
+    - prereq:
```
vutny: I guess we must wrap this with `{%- if postgres.config_backup %} ... {%- endif %}` statements, instead of `onlyif` and `check_cmd`, because no need to trigger backup if it was disabled.

postgres/server/init.sls
```diff
@@ -214,6 +211,18 @@ postgresql-pg_ident:
       - service: postgresql-running
       {%- endif %}
 
+backup_pg_ident:
```
vutny: Like above, we need the `if` statement here. These states also should be applied only when `postgres.config_backup` variable would contain any meaningful value.

vutny:
@thomasrossetto Would you be able to finish this PR and address the comments?
Your changes are great, we just need to resolve the case where local config backup is explicitly disabled. This is long time legacy, and by removing it you will significantly improve the formula.
Thanks for you work and contributions here!

myii:
Good work @thomasrossetto!  I've been working on getting the Travis-based Kitchen tests established for this formula but the `fedora-28` instance was failing.

https://travis-ci.org/myii/postgres-formula/jobs/524280903#L1427:

```salt
                 ID: postgresql-pg_hba
           Function: file.managed
               Name: /var/lib/pgsql/data/pg_hba.conf
             Result: False
            Comment: check_cmd execution failed
```

Applied these commits and it's now working.

https://travis-ci.org/myii/postgres-formula/jobs/524291445#L1439:

```salt
                 ID: postgresql-pg_hba
           Function: file.managed
               Name: /var/lib/pgsql/data/pg_hba.conf
             Result: True
            Comment: File /var/lib/pgsql/data/pg_hba.conf updated
```

While this might not be what your intention was when you submitted this PR, it's a nice bonus.

myii:
@vutny After setting up the CI PR (#262), the `fedora-28` test is now passing, even without this merge.

https://travis-ci.com/saltstack-formulas/postgres-formula/jobs/195597494#L1552:

```salt
                 ID: postgresql-pg_hba
           Function: file.managed
               Name: /var/lib/pgsql/data/pg_hba.conf
             Result: True
            Comment: File /var/lib/pgsql/data/pg_hba.conf updated
```

So no need to merge this PR prematurely, it seems.

vutny:
Great, thank you @myii !

noelmcloughlin:
Thanks, @thomasrossetto @myii  @vutny

PR #257, opened 2019-03-22
Remove deprecated createuser command in salt 2019.02

thomasrossetto:


myii:
@thomasrossetto Thanks for this -- merged.

---

For clarity, quoting the [main documentation for the major version before `createuser` was removed](https://docs.saltstack.com/en/2018.3/ref/states/all/salt.states.postgres_user.html#salt.states.postgres_user.present):

> ```
> createroles
>     Is the user allowed to create other users?
> createuser
>     Alias to create roles
> ```

* I.e. `createroles` should be used instead of `createuser`.

PR #258, opened 2019-04-14
Always retry alternatives.install state

noelmcloughlin:
This PR introduces retries because salt `alternatives.install` sometimes fails on 1st run.

Verified on SuSE

### DIFF
--- a/postgres/client/init.sls
+++ b/postgres/client/init.sls
@@ -39,6 +39,9 @@ postgresql-{{ bin }}-altinstall:
     - onlyif: test -f {{ path }}
     - require:
       - pkg: postgresql-client-libs
+    - retry:
+        attempts: 2
+        until: True
 
     {%- endfor %}
 {%- endif %}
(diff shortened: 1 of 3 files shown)


noelmcloughlin:
Hi @vutny Thanks for reviewing this. The original bug is reported at https://github.com/saltstack/salt/issues/38802  This PR resolves formula error for any salt version so states are marked success not failed.

postgres/client/init.sls
```diff
@@ -39,6 +39,9 @@ postgresql-{{ bin }}-altinstall:
     - onlyif: test -f {{ path }}
     - require:
       - pkg: postgresql-client-libs
+    - retry:
```
vutny: Wrap the argument with:
```jinja
{%- if grains['saltversioninfo'] < [2018, 11, 0, 0] %}
...
{%- endif %}
```

noelmcloughlin: Good suggestion @vutny

noelmcloughlin:
That's done.  I also removed related hack from `postgres/dev/init.sls`

myii:
Thanks for the PR @noelmcloughlin and thanks for taking the time to review @vutny.

PR #261, opened 2019-04-24
fix(uuid-ossp): use hyphen consistently

myii:
* Found when testing Travis CI for the formula
* States fail ~and run in the wrong order~
* Broken: https://travis-ci.org/myii/postgres-formula/jobs/524227136#L2092
* Fixed: https://travis-ci.org/myii/postgres-formula/jobs/524229384#L2086

---

Update: While it definitely fixes things, that's only for `Debian` based images so far.

### DIFF
--- a/pillar.example
+++ b/pillar.example
@@ -174,14 +174,14 @@ postgres:
 
   # optional schemas to enable on database
   schemas:
-    uuid_ossp:
+    uuid-ossp:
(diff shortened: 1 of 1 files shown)


myii:
@vutny Thanks for the review.

myii:
Thanks @aboe76.

PR #265, opened 2019-04-30
fix(postgres.manage): only reload modules if needed; fixes #214

### LINKED ISSUE #214: State postgres-reload-modules always reports changes

alxwr:
Based on the great work of @RobRuana in #216.

Tested on FreeBSD 11.2 with Salt 2019.2.0.

- `postgres-reload-modules` only report changes (and reloads the modules) if necessary
  - uses `test.configurable_test_state`
- `require` is always present in the `format_state` macro, so we're just using it. (This part is not as clean as I want it to be, but I regard it as a trade-off. Maybe there are better ideas.)
- Changing Pillar data in a subsequent highstate run results in changed users, databases, etc., as requested in https://github.com/saltstack-formulas/postgres-formula/pull/216#issuecomment-409503424.

### DIFF
--- a/postgres/defaults.yaml
+++ b/postgres/defaults.yaml
@@ -71,6 +71,8 @@ postgres:
 
   bake_image: False
 
+  manage_force_reload_modules: False
+
   fromrepo: ''
 
   users: {}
--- a/postgres/macros.jinja
+++ b/postgres/macros.jinja
@@ -25,7 +25,7 @@
 {{ state }}-{{ name }}:
   {{ state }}.{{ ensure|default('present') }}:
     {{- format_kwargs(kwarg) }}
-    - onchanges:
+    - require:
       - test: postgres-reload-modules
 
 {%- endmacro %}
--- a/postgres/manage.sls
+++ b/postgres/manage.sls
@@ -1,7 +1,9 @@
 {%- from tpldir + "/map.jinja" import postgres with context -%}
 {%- from tpldir + "/macros.jinja" import format_state with context -%}
 
-{%- if salt['postgres.user_create']|default(none) is not callable %}
+{%- set needs_client_binaries = salt['postgres.user_create']|default(none) is not callable -%}
+
+{%- if needs_client_binaries %}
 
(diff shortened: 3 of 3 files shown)


postgres/defaults.yaml
```diff
@@ -71,6 +71,8 @@ postgres:
 
   bake_image: False
 
+  manage_force_reload_modules: False
```
vutny: No need to have this configurable. We will make the default behavior correct, so there would be no undesired state changes.

postgres/macros.jinja
```diff
@@ -25,7 +25,7 @@
 {{ state }}-{{ name }}:
   {{ state }}.{{ ensure|default('present') }}:
     {{- format_kwargs(kwarg) }}
-    - onchanges:
```
vutny: Instead of making the requirement, if I'd suggest to put a conditional like this:
```jinja
    {%- if needs_client_binaries %}
    - onchanges:
       - test: postgres-reload-modules
    {%- endif %}
```

This would make changes only when the binaries were not available before, which is valid behavior.
Otherwise, there would not be any dependency on the test state, since it would actually do nothing.

postgres/manage.sls
```diff
@@ -18,7 +20,12 @@ include:
 # Ensure that Salt is able to use postgres modules
 
 postgres-reload-modules:
-  test.succeed_with_changes:
+  test.configurable_test_state:
+    - changes:
```
vutny: Let's just make this very simple:
```jinja
    - changes: {{ needs_client_binaries }}
```

Should work as expected.

postgres/manage.sls
```diff
@@ -35,7 +42,7 @@ postgres-reload-modules:
 
 {{ format_state(name, 'postgres_tablespace', tblspace) }}
   {%- if 'owner' in tblspace %}
-    - require:
+{#- - require: #}
```
vutny: Then, all requisite should be returned back.

PR #268, opened 2019-05-13
fix(freebsd-user): fix FreeBSD daemon's user for PostgreSQL >= 9.6

### LINKED ISSUE #263: osFamilyMap.yaml FreeBSD default user
Actually default user for FreeBSD is `pgsql` whereas it is `postgres` on FreeBSD's package installation for Postgres 9.6.
It causes `postgres.server` state to fail without pillar adjustments.

Strade28:
For postgresql 9.6 and newer, the default unix user used by postgresql
daemon has changed to 'postgres', and the default data directory is:
/var/db/postgres/data96
For further details see : [revision 421360](https://svnweb.freebsd.org/ports/head/UPDATING?view=markup&pathrev=421360)

Closes #263

### DIFF
--- a/postgres/osfamilymap.yaml
+++ b/postgres/osfamilymap.yaml
@@ -21,10 +21,17 @@ Debian:
     {% endif %}
 
 FreeBSD:
-  conf_dir: /usr/local/pgsql/data
-  data_dir: /usr/local/pgsql/data
+    {% if repo.version|float >= 9.6 %}
+  user: &freebsd-user postgres
+  group: &freebsd-group postgres
+  conf_dir: {{ '/var/db/postgres/data' ~ release  }}
+  data_dir: {{ '/var/db/postgres/data' ~ release }}
+    {% else %}
   user: &freebsd-user pgsql
   group: &freebsd-group pgsql
+  conf_dir: /usr/local/pgsql/data
+  data_dir: /usr/local/pgsql/data
+    {% endif %}
   pkg_client: postgresql{{ release }}-client
   pkg: postgresql{{ release }}-server
   prepare_cluster:


myii:
@Strade28 Thanks for your PR.  I'm just pinging the relevant people to have a look.

CC: @alxwr @sticky-note.

sticky-note:
@Strade28 @myii Looks good for me

postgres/osfamilymap.yaml
```diff
@@ -21,10 +21,17 @@ Debian:
...
+    {% if repo.version >= 9.6 %}
+  user: &freebsd-user postgres
+  group: &freebsd-group postgres
+  conf_dir: {{ '/var/db/postgres/data' ~ release  }}
+  data_dir: {{ '/var/db/postgres/data' ~ release }}
+    {% else %}
+  user: &freebsd-user {{ pgsql }}
+  group: &freebsd-group {{ pgsql }}
+  conf_dir: {{ '/usr/local/pgsql/data' }}
+  data_dir: {{ '/usr/local/pgsql/data' }}
```
n-rodriguez: why use string interpolation for static strings?

Strade28: No reasons, I  modified

n-rodriguez: thank you!

postgres/osfamilymap.yaml
```diff
@@ -21,10 +21,17 @@ Debian:
     {% endif %}
 
 FreeBSD:
-  conf_dir: /usr/local/pgsql/data
-  data_dir: /usr/local/pgsql/data
+    {% if repo.version >= 9.6 %}
```
vutny: This will not be evaluated correctly, since the `version` value is set as a string in the defaults and the pillar example as well. I suggest to use `float()` filter.
```suggestion
    {% if repo.version|float >= 9.6 %}
```

Strade28: This is done. thank you.

vutny:
Thank you for your contribution, @Strade28

PR #269, opened 2019-05-14
test(locale): improve test using locale `en_US.UTF-8`

myii:
* ~Configure locale on Debian-based instances and CentOS-7~

---

Continues work done in #262, where the locales were commented out at the time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,21 +22,21 @@ env:
     # TODO: Re-enable this once the `systemd` service can be restarted reliably
     # - INSTANCE: default-ubuntu-1804-2019-2-py3
     - INSTANCE: default-centos-7-2019-2-py3
-    - INSTANCE: default-fedora-29-2019-2-py3
+    # - INSTANCE: default-fedora-29-2019-2-py3
     - INSTANCE: default-opensuse-leap-15-2019-2-py3
-    - INSTANCE: default-debian-9-2018-3-py2
+    # - INSTANCE: default-debian-9-2018-3-py2
     - INSTANCE: default-ubuntu-1604-2018-3-py2
-    - INSTANCE: default-centos-7-2018-3-py2
+    # - INSTANCE: default-centos-7-2018-3-py2
     - INSTANCE: default-fedora-29-2018-3-py2
     # TODO: Use this when fixed instead of `opensuse-423`
     # NOTE: Having to temporarily disable `opensuse-423` due to slow mirror
     # - INSTANCE: default-opensuse-leap-15-2018-3-py2
     # - INSTANCE: default-opensuse-423-2018-3-py2
-    - INSTANCE: default-debian-8-2017-7-py2
-    - INSTANCE: default-ubuntu-1604-2017-7-py2
+    # - INSTANCE: default-debian-8-2017-7-py2
+    # - INSTANCE: default-ubuntu-1604-2017-7-py2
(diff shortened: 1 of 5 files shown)


n-rodriguez:
@myii : good work :+1: 

but I have one question : can the locales be set in the docker image instead?

myii:
@vutny Thanks for the feedback.  I've tried it out on a separate branch but I'm hitting the same error on all of the Debian-based instances:

```
       *** update-locale: Error: invalid locale settings:  LANG=en_US.UTF-8
       The command '/bin/sh -c sudo dpkg-reconfigure -f noninteractive locales' returned a non-zero code: 255
```

Full logs available here: https://travis-ci.com/myii/postgres-formula/builds/111870004.

---

In any case, we've discussed this on Slack and @javierbertoli may be able to ensure that we don't need this process at all.  `en_US.UTF-8` is already configured on most of the images, except for `debian-8` and `debian-9`.  If those can be resolved, we can remove all of the `provision_command` sections from this PR.

vutny:
> @vutny Thanks for the feedback. I've tried it out on a separate branch but I'm hitting the same error on all of the Debian-based instances:

Ah, of course! Forgot the very first command :smile: 
```sh
sudo rm -f /etc/default/locale /etc/locale.gen
```

> In any case, we've discussed this on Slack and @javierbertoli may be able to ensure that we don't need this process at all. `en_US.UTF-8` is already configured on most of the images, except for `debian-8` and `debian-9`. If those can be resolved, we can remove all of the `provision_command` sections from this PR.

Yeah, sure. This is the best option. Better to put those commands into Dockerfile.

myii:
@vutny Yes, that worked: https://travis-ci.com/myii/postgres-formula/builds/111883527.  Let's see what the final outcome for `en_US.UTF-8`.

myii:
Thanks to @javierbertoli, all of the pre-salted images are configured with the `en_US.UTF-8` locale.  So all of the `provision_commands` have been removed.

I've also added a couple of further changes based on the `template-formula`.  The first one is to remove the temporary `suse` conditional since that's been fixed upstream (thanks @n-rodriguez!).  The second is to reduce the testing matrix back to the "tree"-shaped matrix.  I did test these changes before reducing the matrix and all was OK:

* https://travis-ci.com/myii/postgres-formula/builds/113817072

@vutny If you are good with the changes, would you mind merging this?

myii:
> Thanks @myii , great work!
> I guess you need to throw away changes in `.travis.yml`, as I see all defined instances have been passed in your other branch covered by Travis.

@vutny It's a bit ugly but we've been leaving them there (commented out) to make it easier when `diff`ing with the `template-formula` (across numerous formulas).  It's especially helpful when there are multiple changes to propagate, as there already have been with `.travis.yml`.  And there are still more changes expected, since we're evaluating other CI solutions.  So I'd appreciate it if we can keep hold of them there, for easier multi-formula maintenance.

myii:
@vutny Sorry, I didn't get your angle about the "temporary thing".  Do you mean in terms of having all of the instances running again?  Or do you mean removing the commented lines?  In any case, if anyone wants to help stabilise the CI setup, that would be greatly appreciated.  Here's a bit more background for interested parties, even if only to offer comment: https://github.com/saltstack-formulas/template-formula/issues/118.

I'll go ahead and merge this in the meanwhile, thanks for the review.

PR #273, opened 2019-07-03
fixes salt-run for kitchen test

fortunejuggle:
fixes kitchen

 this reorders .yml to avoid `pillar failed to render error`

### DIFF
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -118,13 +118,6 @@ provisioner:
     base:
       '*':
         - postgres
-  pillars:
-    top.sls:
-      base:
-        '*':
-          - postgres
-  pillars_from_files:
-    postgres.sls: test/salt/pillar/postgres.sls
 
 verifier:
   # https://www.inspec.io/
@@ -133,8 +126,17 @@ verifier:
   # cli, documentation, html, progress, json, json-min, json-rspec, junit
   reporter:
     - cli
-  inspec_tests:
-    - path: test/integration/default
 
 suites:
   - name: default
+    verifier:
+      inspec_tests:
+        - test/integration/default
+    provisioner:
+      pillars-from-files:
+        postgres.sls: test/salt/pillar/postgres.sls
+      pillars:
+        top.sls:
+          base:
+            '*':
+              - postgres


aboe76:
@fortunejuggle can you please edit your commit message this will fail:

✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]

see docs: https://github.com/saltstack-formulas/postgres-formula/blob/master/docs/CONTRIBUTING.rst

myii:
@fortunejuggle Thanks for providing this PR.  A suggested commit message you can use when amending the commit:

```git
fix(kitchen): reorder `.yml` to avoid `Pillar failed to render` error
```

@javierbertoli @daks Can I borrow your Kitchen expertise to discuss if we need to apply this reordering back in our other formulas as well, starting with the `template-formula`?

fortunejuggle:
finally - it is done - thanks for your help @myii and thanks for stating it @aboe76

daks:
I'm surprised it changes anything. I don't thing there is any notion of order in kitchen. It could exist for salt (like states ordering in defined top.sls) but not for kitchen.

javierbertoli:
@fortunejuggle, I fail to see what's the issue you're trying to fix here. An error `yaml failed to render` means the content of the file is not correct yaml (or that kitchen failed to read it?)

In this case, moving the declarations from the general (default) stanza to the `suites` stanzas does not change a thing regarding the issue you mention, and should not fix it as far as I know. 

Taking into consideration the fact that it 'fixed it', the only difference I see is that, in the stanza, [you used the old/deprecated syntax for `pillars_from_files`](https://github.com/saltstack-formulas/postgres-formula/pull/273/files#diff-fdbca9ed07a634829a3548f874dc925aR136) with dashes instead of underscores. So I wonder if the "real" issue is either an old `kitchen-salt` gem in your testing environment or similar thing, which makes kitchen unable to read the pillar because the variable is 'not declared'? 

OFC I'm guessing here. As I said, the combination of error and solution you propose don't make sense to me.

Can this be a possibility?

daks:
I cloned master and tried to run tests on `default-debian-9-2019-2-py3` and it fails. Are we sure that tests run correctly on standard `salt-call`? And how can `salt-run` be tested?

fortunejuggle:
@daks -the heading is `fixes salt-run for kitchen tests`.
I always test the tests first - that's why I haven't made any progress yet. However, salt ran through the docker container after these changes. However, I have not yet been able to check whether this is generally the case.

fortunejuggle:
@javierbertoli  - the error was that `postgres was not found in enviraonment base`

myii:
Thanks for joining in @daks and @javierbertoli.  I'll try to make some time to do some local tests at my end as well.  @fortunejuggle Thanks for your patience, it's a good chance for us to review the test setup, to see whether anything else can be done from our end to ease local testing.  By the way, can you share how you are calling the tests?

As I'm typing this, you just closed the PR.  Is there any reason for that?

javierbertoli:
@fortunejuggle, the current [`kitchen.yaml` stanzas](https://github.com/saltstack-formulas/postgres-formula/blob/master/kitchen.yml#L121-L140) pass (see https://travis-ci.com/saltstack-formulas/postgres-formula)

I tried the current master of the repo locally and passed, so I can't reproduce your error. 

Is it possible you paste the `kitchen.yaml` and `pillar` you're using to test, as well as the error you get (the logs of it, I mean)

fortunejuggle:
@myii  - sorry did not mention you - I have used the wrong kitchen-salt gem. All other gems are fine. So it was my mistake - @daks can you please also check it again? @javierbertoli - yeah please refer my previous comment . 

I'm trying to be more careful - I'm working a lot with salt right now and I'm trying to help, this time it went wrong.

myii:
@fortunejuggle Not a problem, welcome to Salt!  It's good to know that our testing is working as expected.  Thanks to @daks and @javierbertoli for confirming that with their advice.

myii:
And to @aboe76  for the initial review!

PR #278, opened 2019-07-24
feat(debian): add buster support

### LINKED ISSUE #276: Does not work with Debian Buster
Debian buster fails because it's not added to the yaml files.

johnkeates:
feat(debian): add buster support

This adds support for Debian buster and native package installation. Fixes #276

### DIFF
--- a/postgres/codenamemap.yaml
+++ b/postgres/codenamemap.yaml
@@ -54,12 +54,14 @@
 {{ debian_codename('wheezy', '9.1') }}
 {{ debian_codename('jessie', '9.4') }}
 {{ debian_codename('stretch', '9.6') }}
+{{ debian_codename('buster', '11') }}
 
 # `oscodename` grain has long distro name
 # if `lsb-release` package not installed
 {{ debian_codename('wheezy', '9.1', 'Debian GNU/Linux 7 (wheezy)') }}
 {{ debian_codename('jessie', '9.4', 'Debian GNU/Linux 8 (jessie)') }}
 {{ debian_codename('stretch', '9.6', 'Debian GNU/Linux 9 (stretch)') }}
+{{ debian_codename('buster', '11', 'Debian GNU/Linux 10 (buster)') }}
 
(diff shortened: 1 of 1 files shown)


postgres/codenamemap.yaml
```diff
@@ -54,12 +54,14 @@
...
 {{ debian_codename('jessie', '9.4') }}
 {{ debian_codename('stretch', '9.6') }}
+{{ debian_codename('buster', '11') }}
 
 # `oscodename` grain has long distro name
 # if `lsb-release` package not installed
 {{ debian_codename('wheezy', '9.1', 'Debian GNU/Linux 7 (wheezy)') }}
 {{ debian_codename('jessie', '9.4', 'Debian GNU/Linux 8 (jessie)') }}
 {{ debian_codename('stretch', '9.6', 'Debian GNU/Linux 9 (stretch)') }}
+{{ debian_codename('buster', '11', 'Debian GNU/Linux 10 (buster)') }}
```
aboe76: shouldn't this be '10' ?

johnkeates: No, that's a mistake I made as well :p this adds the PostgreSQL version that exists in the distro repo and doesn't refer to the distrio version at all.

johnkeates:
Travis doesn't seem to like my commit message for some reason. But the output doesn't really give me a pointer to what's wrong:

✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]

johnkeates:
Thanks @noelmcloughlin should be all better now

postgres/codenamemap.yaml
aboe76: thanks for pionting this out.

noelmcloughlin:
> should be all better now

No, it's the commit message you need to change. ;-)

johnkeates:
> > should be all better now
> 
> No, it's the commit message you need to change. ;-)

I really should RTFM :-D

PR #280, opened 2019-08-03
feat(archlinux): add support, fixing rendering errors

noelmcloughlin:
There is no `osfinger` grain on Archlinux.

### DIFF
--- a/postgres/upstream.sls
+++ b/postgres/upstream.sls
@@ -39,6 +39,6 @@ postgresql-repo:
 postgresql-repo:
   test.show_notification:
     - text: |
-        PostgreSQL does not provide package repository for {{ grains['osfinger'] }}
+        PostgreSQL does not provide package repository for {{ salt['grains.get']('osfinger', grains.os) }}
 
 {%- endif %}
--- a/test/salt/pillar/postgres.sls
+++ b/test/salt/pillar/postgres.sls
@@ -1,7 +1,7 @@
 # Port to use for the cluster -- can be used to provide a non-standard port
 # NOTE: If already set in the minion config, that value takes priority
 
-{%- if not (grains.os_family == 'Debian' or grains.osfinger == 'Leap-15') %}
+{%- if grains.os_family != 'Debian' and salt['grains.get']('osfinger') != 'Leap-15' %}
 postgres.port: '5432'
 {%- else %}
 postgres.port: '5433'
@@ -10,7 +10,7 @@ postgres.port: '5433'
 postgres:
   # UPSTREAM REPO
   # Set True to configure upstream postgresql.org repository for YUM/APT/ZYPP
-  {%- if not (grains.os_family == 'Debian' or grains.osfinger == 'CentOS-6') %}
+  {%- if grains.os_family != 'Debian' and salt['grains.get']('osfinger') != 'CentOS-6' %}
   use_upstream_repo: False
   {%- else %}
   use_upstream_repo: True
@@ -181,7 +181,7 @@ postgres:
         public:
           owner: localUser
       # enable per-db extension
-      {%- if grains.os_family == 'Debian' and grains.osfinger != 'Debian-8' %}
+      {%- if grains.os_family == 'Debian' and salt['grains.get']('osfinger') != 'Debian-8' %}
       extensions:
         uuid-ossp:
           schema: 'public'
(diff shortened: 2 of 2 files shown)


postgres/upstream.sls
```diff
@@ -39,6 +39,6 @@ postgresql-repo:
 postgresql-repo:
   test.show_notification:
     - text: |
-        PostgreSQL does not provide package repository for {{ grains['osfinger'] }}
+        PostgreSQL does not provide package repository for {{ grains['os'] }}
```
myii: Does this work?

```suggestion
        PostgreSQL does not provide package repository for {{ grains['osfinger'] | d(grains['os']) }}
```

That gives us the best of both worlds.

noelmcloughlin: It's a jinja rendering error so that will not fix.
 If we really want osfinger then will need a if/else block

vutny: I think calling `salt['grains.get']('osfinger', grains['os'])` would be a better choice here. No need for additional conditional statements.
Since that is just a notification message, having exactly `osfinger` value substituted is not such critical matter.

noelmcloughlin: Thanks @myii and @vutny for feedback.  I have updated as suggested by @vutny and since `salt['grains.get']` swallows the exception this should work in all cases.

noelmcloughlin:
One travis CI job failure https://github.com/saltstack-formulas/postgres-formula/issues/282 is unrelated to this PR.

test/salt/pillar/postgres.sls
```diff
@@ -1,7 +1,7 @@
 # Port to use for the cluster -- can be used to provide a non-standard port
 # NOTE: If already set in the minion config, that value takes priority
 
-{%- if not (grains.os_family == 'Debian' or grains.osfinger == 'Leap-15') %}
+{%- if not (grains.os_family == 'Debian' or ('finger' in grains and grains.osfinger == 'Leap-15')) %}
```
vutny: I guess the string should be `osfinger` instead `finger`.
But why so complicated? We could use the same `salt['grains.get']()` function in Pillar file as well.

noelmcloughlin: Good catch - thanks.

noelmcloughlin:
Thanks @vutny fixed up the commit now.

test/salt/pillar/postgres.sls
```diff
@@ -1,7 +1,7 @@
 # Port to use for the cluster -- can be used to provide a non-standard port
 # NOTE: If already set in the minion config, that value takes priority
 
-{%- if not (grains.os_family == 'Debian' or grains.osfinger == 'Leap-15') %}
+{%- if not (grains.os_family == 'Debian' or salt['grains.get']('osfinger') == 'Leap-15')) %}
```
vutny: CI fails because of too many parentheses, @noelmcloughlin :sweat_smile:

noelmcloughlin: there's always one!!

test/salt/pillar/postgres.sls
```diff
@@ -181,7 +181,7 @@ postgres:
         public:
           owner: localUser
       # enable per-db extension
-      {%- if grains.os_family == 'Debian' and grains.osfinger != 'Debian-8' %}
+      {%- if grains.os_family == 'Debian' and salt['grains.get']('osfinger') != 'Debian-8') %}
```
vutny: This one at the end and later are superfluous too :smiley:

noelmcloughlin: I nailed them all now. Cmon *Lint* !!

test/salt/pillar/postgres.sls
```diff
@@ -1,7 +1,7 @@
 # Port to use for the cluster -- can be used to provide a non-standard port
 # NOTE: If already set in the minion config, that value takes priority
 
-{%- if not (grains.os_family == 'Debian' or grains.osfinger == 'Leap-15') %}
+{%- if not (grains.os_family == 'Debian' or salt['grains.get']('osfinger') == 'Leap-15' %}
```
vutny: Here the parentheses were valid. But we could remove them by simplifying Boolean logic:
`not (var == val or var == val)` is the same as `var != val or var != val`.

noelmcloughlin: No problem - that's done with updated logic.

PR #289, opened 2019-09-10
fix(freebsd): no libpqdev freebsd package

noelmcloughlin:
Resolve failed state on FreeBSD.
```
 - pkg: No packages available to install matching 'libpq-dev' 
              have been found in the repositories
```

### DIFF
--- a/postgres/osfamilymap.yaml
+++ b/postgres/osfamilymap.yaml
@@ -35,6 +35,7 @@ FreeBSD:
   conf_dir: /usr/local/pgsql/data
   data_dir: /usr/local/pgsql/data
     {% endif %}
+  pkg_libpq_dev: null  #not in freebsd
   pkg_client: postgresql{{ release }}-client
   pkg: postgresql{{ release }}-server
   prepare_cluster:


postgres/osfamilymap.yaml
```diff
@@ -35,6 +35,7 @@ FreeBSD:
   conf_dir: /usr/local/pgsql/data
   data_dir: /usr/local/pgsql/data
     {% endif %}
+  pkg_libpq_dev:
```
n-rodriguez: I don't if the new linter config is merged but you should set it to `null` if possible
cc @myii

myii: Nice catch, @n-rodriguez.  I've just pushed the PRs for `empty-values` to 40 of our repos (see https://github.com/saltstack-formulas/template-formula/issues/164).  So yes, each value should be set explicitly.  It's easiest to use `~` to represent `null`, if that's what's required here.  Or even an empty string `''` if that's what's appropriate.

myii: #287 here.

noelmcloughlin: Thanks guys! Great catch ...

So `pkg_libpq_dev: ~` is same as `pkg_libpq_dev: Null` ? This is new to me.

myii: @noelmcloughlin As linked in https://github.com/saltstack-formulas/template-formula/issues/164:

* https://yaml.org/type/null.html

I.e. `pkg_libpq_dev:` without any value after it is _implicitly_ `null`.  So it's no different to `pkg_libpq_dev: ~`, except that we're setting it _explicitly_, avoiding any confusion about what is intended.  This then allows us to ensure there are never any empty values in our YAML, which will be in enforced with the merge of #287.

noelmcloughlin: done. Thanks @myii , @n-rodriguez

n-rodriguez: > I.e. `pkg_libpq_dev:` without any value after it is _implicitly_ `null`. So it's no different to `pkg_libpq_dev: ~`,

If the result is the same, I prefer `null` over `~` to avoid confusion with Jinja concatenation operator.

noelmcloughlin: I agree - done.

PR #309, opened 2021-02-25
fix(repo): reorder dependencies to prevent errors

javierbertoli:
### PR progress checklist (to be filled in by reviewers)

- [ ] Changes to documentation are appropriate (or tick if not required)
- [ ] Changes to tests are appropriate (or tick if not required)
- [ ] Reviews completed

---

#### Primary type

- [ ] `[build]`    Changes related to the build system
- [ ] `[chore]`    Changes to the build process or auxiliary tools and libraries such as documentation generation
- [ ] `[ci]`       Changes to the continuous integration configuration
- [ ] `[feat]`     A new feature
- [X] `[fix]`      A bug fix
- [ ] `[perf]`     A code change that improves performance
- [ ] `[refactor]` A code change that neither fixes a bug nor adds a feature
- [ ] `[revert]`   A change used to revert a previous commit
- [ ] `[style]`    Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc.)

#### Secondary type

- [ ] `[docs]`     Documentation changes
- [ ] `[test]`     Adding missing or correcting existing tests

### Does this PR introduce a `BREAKING CHANGE`?

No.

### Describe the changes you're proposing

When installing a `postgresql-client` using upstream's repo, the formula fails to install the repo because a `require_in` dependency expects the `postgresql-server` package. This PR moves the dependency to the `server/init.sls` state.
 
### Pillar / config required to test the proposed changes

The issue can be reproduced using this pillar
```yaml
postgres:
  use_upstream_repo: true
  version: 11
```
and this command
```
salt-call --local state.apply postgres.client
```
### Documentation checklist

- [ ] Updated the `README` (e.g. `Available states`).
- [ ] Updated `pillar.example`.

### Testing checklist

- [ ] Included in Kitchen (i.e. under `state_top`).
- [ ] Covered by new/existing tests (e.g. InSpec, Serverspec, etc.).
- [ ] Updated the relevant test pillar.

### DIFF
--- a/postgres/client/init.sls
+++ b/postgres/client/init.sls
@@ -16,7 +16,7 @@ include:
 postgresql-client-libs:
   pkg.installed:
     - pkgs: {{ pkgs | json }}
-  {%- if postgres.use_upstream_repo == true %}
+  {%- if 'pkg_repo' in postgres and postgres.use_upstream_repo == true %}
     - refresh: True
     - require:
       - pkgrepo: postgresql-repo
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -18,7 +18,7 @@ include:
 postgresql-server:
   pkg.installed:
     - pkgs: {{ pkgs | json }}
-{%- if postgres.use_upstream_repo == true %}
+{%- if 'pkg_repo' in postgres and postgres.use_upstream_repo == true %}
     - refresh: True
     - require:
       - pkgrepo: postgresql-repo
--- a/postgres/server/init.sls
+++ b/postgres/server/init.sls
@@ -20,6 +20,8 @@ postgresql-server:
     - pkgs: {{ pkgs | json }}
 {%- if postgres.use_upstream_repo == true %}
     - refresh: True
+    - require:
+      - pkgrepo: postgresql-repo
 {%- endif %}
   {%- if postgres.fromrepo %}
     - fromrepo: {{ postgres.fromrepo }}
--- a/postgres/python.sls
+++ b/postgres/python.sls
@@ -9,7 +9,7 @@ postgresql-python:
   {% if postgres.fromrepo %}
     - fromrepo: {{ postgres.fromrepo }}
   {% endif %}
-  {% if postgres.use_upstream_repo == true %}
+  {% if 'pkg_repo' in postgres and postgres.use_upstream_repo == true %}
     - refresh: True
     - require:
       - pkgrepo: postgresql-repo
--- a/postgres/upstream.sls
+++ b/postgres/upstream.sls
@@ -28,8 +28,6 @@ postgresql-repo:
     {{- format_kwargs(postgres.pkg_repo) }}
     - require:
       - pkg: postgresql-pkg-deps
-    - require_in:
-      - pkg: postgresql-server
 
   {%- else -%}
 


sticky-note:
This is the exact contrary of this https://github.com/saltstack-formulas/postgres-formula/pull/300
which was introduced because when `use_upstream_repo: True` on `FreeBSD` the `require:  postgresql-repo` fails to find the right due probably to this line https://github.com/saltstack-formulas/postgres-formula/blob/7d46b273af8ae46b2e2c9bfec41b417c6eeb24aa/postgres/upstream.sls#L4

Not sure of what I am saying, I had the exact same  problem today but this PR need to be double checked

postgres/server/init.sls
```diff
@@ -20,6 +20,8 @@ postgresql-server:
     - pkgs: {{ pkgs | json }}
 {%- if postgres.use_upstream_repo == true %}
```
vutny, line 21: This statement probably needs additional expression, the same as in `upstream.sls`:
```suggestion
{%- if 'pkg_repo' in postgres and postgres.use_upstream_repo == true %}
```
Because if "upstream" is flag is on, but there are no repo definition for particular distro, the requisite will fail.

vutny (changes requested):

postgres/server/init.sls
javierbertoli: @sticky-note, I added the changes @vutny suggests and also added the same check in a few other places where `postgres.use_upstream_repo` was being checked but `pkg_repo` was not.

I think the changes in this PR should be OK for all the platforms, but I don't have a _FreeBSD_ around to try it.  Can you try it and give us your feedback?

postgres/client/init.sls
```diff
@@ -7,7 +7,7 @@
   {%- endif %}
 {%- endfor %}
 
-{%- if postgres.use_upstream_repo == true %}
+{%- if 'pkg_repo' in postgres and postgres.use_upstream_repo == true %}
```
vutny, line 10: Here, it is not strictly required, since underlying logic in `postgres.upstream` will handle it anyway.
But does no harm as well.

postgres/server/init.sls
```diff
@@ -4,7 +4,7 @@
 {%- if postgres.bake_image %}
   {%- do includes.append('postgres.server.image') %}
 {%- endif %}
-{%- if postgres.use_upstream_repo == true -%}
+{%- if 'pkg_repo' in postgres and postgres.use_upstream_repo == true -%}
```
vutny, line 7: The same as above.

vutny (changes requested):
Thanks @javierbertoli , I just have minor comment about MacOS case.

postgres/dev/init.sls
```diff
@@ -49,7 +49,7 @@ postgres_maxfiles_limits_conf:
     - group: {{ postgres.group }}
   {% endif %}
 
-  {% if postgres.use_upstream_repo == 'postgresapp' %}
+  {% if 'pkg_repo' in postgres and postgres.use_upstream_repo == 'postgresapp' %}
```
vutny, line 52: This might break MacOS installation, there is a special case of `postgres.use_upstream_repo` value.
The `postgres.pkg_repo` is not defined for it, as I recall. And the `upstream.sls` is not included here.

postgres/client/init.sls
javierbertoli: Yes, I know. Just thought that being consistent in the checks within the state file would make it easier to follow the logic.

I can remove those checks if you prefer and just leave the strictly required ones.

postgres/dev/init.sls
javierbertoli: Gotcha. Removed the check.

postgres/client/init.sls
vutny: I mention this only beacase the `upstream.sls` has the notification state, which explicitly shows message about additional repository will not be configured.
This was done for some clarity, then the `postgres.use_upstream_repo` was set, but actually will not going to have any effect.

javierbertoli: So, would you like me to remove the check? :smile:

vutny: I think it would be better, but up to you :smile:

javierbertoli: Done

vutny (approved):
Looks good to me, thank you @javierbertoli !

sticky-note:
Ok, it seems to have no issue in this manner. Thans @javierbertoli

javierbertoli:
Thanks to you both!

PR #314, opened 2021-07-06
fix(osmap.yaml): fix PGDG repo file creation for Amazon Linux 2

Niyakiy:
Fix improper pgdg repo file creation for Amazon Linux 2.

Description:
Current formula implementation uses 
`baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/redhat/rhel-$releasever-$basearch'`
setting $releasever to 2 for Amazon Linux 2 and making formula install fail due to missing rhel-2 packages.

The fix makes a newly created formula repo file to use rhel-7 as the proper repository to use during install.

### DIFF
--- a/postgres/osmap.yaml
+++ b/postgres/osmap.yaml
@@ -9,3 +9,7 @@ Fedora:
     baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/fedora/fedora-$releasever-$basearch'
   remove:
     releases: ['9.4', '9.5', '9.6', '10']
+
+Amazon:
+  pkg_repo:
+    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/redhat/rhel-7-$basearch'


postgres/osfamilymap.yaml
```diff
@@ -61,7 +61,11 @@ RedHat:
     humanname: PostgreSQL {{ repo.version }} $releasever - $basearch
     gpgcheck: 1
     gpgkey: 'https://download.postgresql.org/pub/repos/yum/RPM-GPG-KEY-PGDG-{{ release }}'
+{%- if grains.get('osfullname', "") == "Amazon Linux" and grains.get('osmajorrelease', 0) == 2 %}
+    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/redhat/rhel-7-$basearch'
+{%- else %}
     baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/redhat/rhel-$releasever-$basearch'
+{%- endif %}
```
myii, line 68: @Niyakiy This `if` block can be avoided by adding this change to `osmap.yaml` instead, in a similar manner to this block:

https://github.com/saltstack-formulas/postgres-formula/blob/43c56966e784995248808d48d38eee6b8af00b3a/postgres/osmap.yaml#L6-L9

I.e. Probably:

```yaml
Amazon:
  pkg_repo:
    baseurl: 'https://download.postgresql.org/pub/repos/yum/{{ repo.version }}/redhat/rhel-7-$basearch'
```

This isn't _perfect_, since it doesn't specify the version of Amazon Linux but it is probably sufficient, since we're no longer supporting Amazon Linux 1 (2018).  We _could_ use the `oscodename` grain (`Amazon Linux 2`) in `codenamemap.yaml` but that's overkill in my opinion.  Ultimately, we have a new `map.jinja` that is being introduced across formulas and we can provide the best solution when that becomes available.

myii (changes requested):
Thanks for the PR, @Niyakiy.  Please see the requested change below.

myii:
Just another note, doesn't necessarily have to be addressed by this PR.

---

We're not currently testing `use_upstream_repo` for most platforms, so this PR isn't actually being tested properly.  Need to consider updating this block in the test pillar:

https://github.com/saltstack-formulas/postgres-formula/blob/d16e91ceb03411f58c98cbb919f65b3f3eb4abbb/test/salt/pillar/postgres.sls#L10-L18

I'll try to run some tests when I get a chance.  Let's not make this a blocker for this PR, if I don't report back soon.

@Niyakiy Please don't go out of your way to run these tests, what you've already provided is more than enough, thanks.

postgres/osfamilymap.yaml
Niyakiy: @myii Thanks for the heads up on a better and easier way to introduce a fix. Having AL1 already deprecated it looks reasonable to leave `Amazon` inside `osmap.yaml` as we have only the 2nd version available. However, it will require additional code changes here once AWS releases a new major version of AL. 
I'm about to push the requested code changes along with reworded commit message.

myii (approved):

myii:
Thanks, @Niyakiy -- merged.

Out of interest, did you need to adjust the `pkg_deps` to also install these packages?

```yaml
  pkgs_deps:
    - libicu
    - systemd-sysv
```

I had to do that to get PostgreSQL 13 installed on Amazon Linux 2:

* https://gitlab.com/myii/postgres-formula/-/jobs/1402800380

myii:
@Niyakiy I've refactored this Amazon Linux 2 support in #315 -- feel free to comment on it there.

